Issue #7, opened 2016-09-28
Drop JSON configs in favor of YAML


JackuB:
My reason would be that JSON is more machine readable and we want these configs to be easier to audit. Compare this part of config:

```json
{
  "public":
  [
      {
        "//": "used for pushing up webhooks from github",
        "method": "POST",
        "path": "/webhook/github",
        "valid": [
          {
            "//": "accept all pull request state changes (these don't have files in them)",
            "path": "pull_request.state",
            "value": "open"
          },
          {
            "path": "commits.*.added.*",
            "value": "package.json"
          },
          {
            "path": "commits.*.modified.*",
            "value": "package.json"
          },
          {
            "path": "commits.*.added.*",
            "value": ".snyk"
          },
          {
            "path": "commits.*.modified.*",
            "value": ".snyk"
          }
        ]
      }
  ]
}
```

```yaml
public:
# used for pushing up webhooks from github
- method: POST
  path: "/webhook/github"
  valid:
  # accept all pull request state changes (these don't have files in them)
  - path: pull_request.state
    value: open
  - path: commits.*.added.*
    value: package.json
  - path: commits.*.modified.*
    value: package.json
  - path: commits.*.added.*
    value: ".snyk"
  - path: commits.*.modified.*
    value: ".snyk"
```

Thoughts?

guypod:
No objection here, though I don't feel it's much of a priority. 
If you wanted to do a PR, we wouldn't object.

JackuB:
Resolved by #10

PR #10, opened 2016-10-07
feat: support yaml config

iameugenejo:
- [x] Ready for review
- [x] Reviewed by @gjvis

#### What does this PR do?
It supports yaml config parsing in addition to js/json config parsing

`lib/index.js`

- `npm install` to install js-yaml
- copy and convert existing json config into yaml config
- `npm start` and observe the config being loaded

Commenting is allowed in yaml files which is a huge advantage over using json files.

#7 

#### Screenshots
N/A

#### Additional questions
N/A

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -36,11 +36,7 @@ function main({ port, client } = {}) {
     debug('loading rules from %s', config.accept);
     const acceptLocation = path.resolve(process.cwd(), config.accept);
 
-    if (acceptLocation.endsWith(".yaml") || acceptLocation.endsWith(".yml")) {
-        filters = yaml.safeLoad(fs.readFileSync(acceptLocation, 'utf8'))
-    } else {
-        filters = require(acceptLocation);
-    }
+    filters = yaml.safeLoad(fs.readFileSync(acceptLocation, 'utf8'));
   }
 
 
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -32,7 +34,13 @@ function main({ port, client } = {}) {
...
   if (config.accept) {
     debug('loading rules from %s', config.accept);
-    filters = require(path.resolve(process.cwd(), config.accept));
+    const acceptLocation = path.resolve(process.cwd(), config.accept);
+
+    if (acceptLocation.endsWith(".yaml") || acceptLocation.endsWith(".yml")) {
+        filters = yaml.safeLoad(fs.readFileSync(acceptLocation, 'utf8'))
+    } else {
+        filters = require(acceptLocation);
+    }
```
gjvis: YAML is a superset of JSON (i.e. valid JSON is valid YAML), so we don't need to check file extensions – reading the file with `yaml.safeLoad` will work for both JSON and YAML.

joshje: Those are valid points. 👍 

Our thinking here is that being more strict about what we accept is a good thing. It will reduce the chances of things that could go wrong in the loading of the config.

Removing this logic statement here would be another win as it would simplify the code, making it easier to understand what is accepted (anything that is valid YAML, including valid JSON).

iameugenejo: Understood, another commit is pushed for it.

PR #12, opened 2016-10-11
Support internal CA certificates

### LINKED ISSUE #8: How to handle self-signed certs on Internal services that Client interacts with
The best option for these cases seems to have an envvar with a CA certificate. I'd discourage anyone from using it with `NODE_TLS_REJECT_UNAUTHORIZED` - and let's see if we even get to such situation. However, if this happens, maybe we can improve it with:

- certificate pinning (Client-Server) - not sure if this can be enforced with `NODE_TLS_REJECT_UNAUTHORIZED` active
- Additional encryption/TLS? MAC signing?

I'd remove the `NODE_TLS_REJECT_UNAUTHORIZED` note from Readme as it is too dangerous in my opinion.

JackuB:
- [x] Ready for review
- [x] Reviewed by @gjvis

#### What does this PR do?
- Adds an option to specify a CA cert on Client
- Removes `NODE_TLS_REJECT_UNAUTHORIZED` note

`test/functional/client-ca.test.js` shows usage of this option.

Running Client connecting to a self-signed cert. You could use those in `test/fixtures/certs` (Common Name is set to `localhost`)

One improvement could be to allow the CA file to be embedded directly into the envvar, instead of a file, but let's see if that'll be needed.

I left `test/fixtures/certs` with all of the files, which we can use for different scenarios in the future.

Resolves #8


lib/relay.js
```diff
@@ -112,6 +114,9 @@ function responseHandler(filterRules, config) {
         headers: headers,
         method,
         body,
+        agentOptions: {
+          ca: (() => {return config.caCert ? fs.readFileSync(path.resolve(process.cwd(), config.caCert)) : undefined;})(),
```
gjvis: This is going to read the config file on every request right? I think we should just read it once in `/lib/config/index.js`, and set `config.caCert` to the contents of the file. A CA cert is small enough that the memory overhead doesn't matter. What do you think?

JackuB: Good idea, missed that this will get reevaluated on request, I'll push an update.

PR #18, opened 2017-03-09
Fix/fix ws vuln

aviadatsnyk:
- [x] Ready for review

#### What does this PR do?
fix a vuln in this package, according to the snyk wizard.


.snyk
```diff
@@ -0,0 +1,4 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.7.0
+ignore: {}
+patch: {}
```
gjvis: doesn't look like we actually need this file?

package.json
```diff
@@ -46,18 +46,18 @@
...
-    "engine.io-client": "Snyk/engine.io-client#1.6.11-patched",
+    "engine.io": "^1.8.3",
+    "engine.io-client": "^1.8.3",
     "express": "^4.14.0",
+    "js-yaml": "^3.6.1",
     "path-to-regexp": "^1.5.3",
     "primus": "^6.0.1",
     "primus-emitter": "^3.1.1",
     "request": "^2.74.0",
+    "snyk": "^1.19.1",
```
gjvis: doesn't look like `snyk` is needed?

.snyk
aviadatsnyk: removed

PR #21, opened 2017-03-22
feat: add clients' health endpoint

aviadatsnyk:
- [x] Ready for review

#### What does this PR do?
adds the client's connection status to the server healthcheck.

run both a server and a client locally. call `/healthcheck` before and after stopping the client, and then restart the client again and call `/healthcheck` again.

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -15,8 +15,19 @@ module.exports = ({ config = {}, port = null, filters = {} }) => {
     config,
   });
 
-  app.all('/broker/:id/*', (req, res, next) => {
+  // before the `app.all('/broker/:id/*'` route
+  app.get('/broker/:id', (req, res) => {
+    const id = req.params.id;
 
+    if (connections.has(id)) {
+      return res.status(200).json({ ok: true });
+    }
+
+    debug('no broker found matching "%s"', id);
+    return res.status(404).json({ ok: false });
+  });
+
+  app.all('/broker/:id/*', (req, res, next) => {
     const id = req.params.id;
 
     // check if we have this broker in the connections
@@ -34,6 +45,9 @@ module.exports = ({ config = {}, port = null, filters = {} }) => {
 
     next();
   }, relay.request(filters.public));
+
+  app.get('/healthcheck', (req, res) =>
+          res.status(200).json({ ok: true }));
 
   return {
     io,
--- a/lib/server/socket.js
+++ b/lib/server/socket.js
@@ -25,6 +25,7 @@ module.exports = ({ server, filters, config }) => {
           return connections.set(token, clientPool);
         }
         debug('removing client %s', token);
+        // TODO: log/emit warning
         return connections.delete(token);
       }
     };
--- a/test/functional/healthcheck.test.js
+++ b/test/functional/healthcheck.test.js
@@ -12,6 +12,8 @@ test('proxy requests originating from behind the broker client', t => {
    * 1. start broker in server mode
    * 2. start broker in client mode and join (1)
    * 3. check /healthcheck on client and server
+   * 4. stop client and check it's on "disconnected" in the server
+   * 5. restart client with same token, make sure it's not in "disconnected"
    */
 
   process.env.ACCEPT = 'filters.json';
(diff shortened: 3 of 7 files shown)


lib/server/socket.js
```diff
@@ -25,6 +25,10 @@ module.exports = ({ server, filters, config }) => {
           return connections.set(token, clientPool);
         }
         debug('removing client %s', token);
+        // TODO: log/emit warning
+        connections.set(
+          'disconnected',
+          (connections.get('disconnected') || new Set()).add(token));
```
gjvis: I don't think this should be tracked in the same `Map` that holds actual sockets – too much potential for confusion.

lib/server/index.js
```diff
@@ -44,3 +58,8 @@ module.exports = ({ config = {}, port = null, filters = {} }) => {
     },
   };
 };
+
+
+function hash(s) {
+  return 'sha512::' + createHash('sha512').update(s).digest('hex');
```
adrukh: Isn't creating a hash object somewhat computationally heavy? Why not create one on startup and use it?

gjvis: should we use an environmentally provided salt?

adrukh:
With regards to client id hashing, I'd like to vote *not to hash*. The purpose of this feature is to identify which

adrukh:
Can you move the `disconnected` map outside of the `connections` map? A separate global object is better than extending the purpose of the existing `connections` map.

lib/server/index.js
```diff
@@ -35,6 +35,18 @@ module.exports = ({ config = {}, port = null, filters = {} }) => {
     next();
   }, relay.request(filters.public));
 
+  app.get('/healthcheck', (req, res, next) =>
+          res.status(200).json({ ok: true }));
```
adrukh: No need for `next` arg here, remove for brevity

lib/server/index.js
```diff
@@ -35,6 +35,18 @@ module.exports = ({ config = {}, port = null, filters = {} }) => {
...
+
+  // get status for a client.
+  // need the `connections` in scope, so defined here.
+  app.get('/client-health/:id', (req, res, next) => {
+    const id = req.params.id;
+    return res.status(200).json({
+      ok: !connections.get('DISCONNECTED').has(id)
+    });
+  });
```
adrukh: No need for `next` arg here, remove for brevity

test/functional/healthcheck.test.js
```diff
@@ -55,13 +60,52 @@ test('proxy requests originating from behind the broker client', t => {
...
+        client.close();
+        setTimeout(() => {
+          request({url: serverClientHealth, json: true }, (err, res) => {
+            if (err) { return t.threw(err); }
+
+            t.equal(res.statusCode, 200, '200 statusCode');
+            t.equal(res.body['ok'], false, '{ ok: false } in body');
+            t.end();
+          });
+        }, 100);
```
gjvis: Perhaps we could use the `close` event instead of waiting with setTimeout?

```javascript
client.io.on('close', () => {
  request({url: serverClientHealth, json: true }, (err, res) => {
    if (err) { return t.threw(err); }
    t.equal(res.statusCode, 200, '200 statusCode');
    t.equal(res.body['ok'], false, '{ ok: false } in body');
    t.end();
  });
});
```

If the socket never closes then we'd get slower failure, but success would be faster at least.

(i'm not sure this would actually work, just an idea)

test/functional/healthcheck.test.js
```diff
@@ -55,13 +60,52 @@ test('proxy requests originating from behind the broker client', t => {
...
+
       t.test('clean up', t => {
         client.close();
+
         setTimeout(() => {
           server.close();
           t.ok('sockets closed');
           t.end();
-        }, 100);
+        }, 300);
```
gjvis: this is interesting – I wonder what's caused the shutdown to take longer...

aviadatsnyk: just the 2 tests above. I wanted some more time for the to finish. This might work just as well with the old `100`

test/functional/healthcheck.test.js
aviadatsnyk: I did not pursue this so far.

test/functional/healthcheck.test.js
aviadatsnyk: reduced times again, utilising the serial nature of tap.

PR #25, opened 2017-04-02
feat: proxy broker according to https_proxy and no_proxy environment …

aviadatsnyk:
…variables

- [x] Ready for review

#### What does this PR do?
proxies the connection from the broker client to the server through a proxy, defined in the `https_proxy` and `no_proxy` environment variables.

### DIFF
--- a/lib/patch-https-request-for-proxying.js
+++ b/lib/patch-https-request-for-proxying.js
@@ -1,0 +1,102 @@
+/*
+ * Monkey-patch https.request to proxy request to the broker server
+ * according the the env variables `https_proxy` and `no_proxy`:
+ * get the https proxy from the env. proxy requests unless the
+ * broker server is in the no_proxy list.
+ * Start from the "Entry point" below.
+ */
+const url = require('url');
+const tunnel = require('tunnel');
+const {brokerServerUrl, httpsProxy, noProxy} = require('./config');
+const brokerServer = url.parse(brokerServerUrl || '');
+brokerServer.port = brokerServer.port ||
+  brokerServer.protocol === 'https:' ? '443' : '80';
+
+// adapted from https://github.com/request/request/master/lib/getProxyFromURI.js
+
+function formatHostname(hostname) {
+  // canonicalize the hostname, so that 'oogle.com' won't match 'google.com'
+  return hostname.replace(/^\.*/, '.').toLowerCase();
+}
+
+function parseNoProxyZone(zone) {
+  zone = zone.trim().toLowerCase();
+
+  const zoneParts = zone.split(':', 2);
+  const zoneHost = formatHostname(zoneParts[0]);
+  const zonePort = zoneParts[1];
+  const hasPort = zone.indexOf(':') > -1;
+
+  return {hostname: zoneHost, port: zonePort, hasPort: hasPort};
+}
+
+function uriInNoProxy(uri, noProxy) {
+  const port = uri.port || (uri.protocol === 'https:' ? '443' : '80');
+  const hostname = formatHostname(uri.hostname);
+  const noProxyList = noProxy.split(',');
+
+  // iterate through the noProxyList until it finds a match.
+  return noProxyList.map(parseNoProxyZone).some(function (noProxyZone) {
+    const isMatchedAt = hostname.indexOf(noProxyZone.hostname);
+    const hostnameMatched = (
+      isMatchedAt > -1 &&
+        (isMatchedAt === hostname.length - noProxyZone.hostname.length));
+
+    if (noProxyZone.hasPort) {
+      return (port === noProxyZone.port) && hostnameMatched;
+    }
+
+    return hostnameMatched;
+  });
+}
+
+function shouldProxy(uri) {
+  // Decide the proper request proxy to use based on the request URI object and the
+  // environmental variables (NO_PROXY, HTTP_PROXY, etc.)
+  // respect NO_PROXY environment variables (see: http://lynx.isc.org/current/breakout/lynx_help/keystrokes/environments.html)
+
+  // if no https proxy is defined - don't proxy
+
+  if (!httpsProxy) {
+    return false;
+  }
+
+  // if the noProxy is a wildcard then return null
+
+  if (noProxy === '*') {
+    return false;
+  }
+
+  // if the noProxy is not empty and the uri is found return null
+
+  if (noProxy && noProxy !== '' && uriInNoProxy(uri, noProxy)) {
+    return false;
+  }
+
+  // we should proxy
+  return true;
+}
+
+// Entry point: To patch or not to patch?
+if (brokerServer.host && httpsProxy && shouldProxy(brokerServer)) {
+  const proxy = url.parse(httpsProxy);
+  const tunnelingAgent = tunnel.httpsOverHttp({
+    proxy: {
+      host: proxy.hostname,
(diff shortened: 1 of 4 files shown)


lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,104 @@
...
+// Entry point: To patch or not to patch?
+if (httpsProxy && shouldProxy(brokerServer)) {
+  const proxy = url.parse(httpsProxy);
+
+  let Tunnel;
+  // Tunnel = tunnel.httpsOverHttps;
+  // if (proxy.protocol === 'http:' || (!proxy.protocol && proxy.port === '80')) {
+  Tunnel = tunnel.httpsOverHttp;
+  // }
+  const tunnelingAgent = Tunnel({
```
adrukh: This can be cleaner - `const tunnelingAgent = tunnel.httpsOverHttp({...`

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  });
+
+  // actual monkey patching: BEWARE!
+  const https = require('https');
+  const oldhttpsreq = https.request;
+  https.request = function (options, callback) {
+    if (options.host === brokerServer.host &&
+        // use == not === to match string and int due to options diffs
+        (options.port && options.port == brokerServer.port)) {
```
adrukh: This code means "don't proxy if `options.port` is missing" - do we want that? How about:
```
if (options.host === brokerServer.host) {
  if (options.port == brokerServer.port || (!options.port && brokerServer.port == 80)) {
    ...
```

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,104 @@
...
+ * broker server is in the no_proxy list.
+ * Start from the "Entry point" below.
+ */
+const url = require('url');
+const tunnel = require('tunnel');
+const {brokerServerUrl} = require('./config');
+const brokerServer = url.parse(brokerServerUrl || '');
+brokerServer.port = brokerServer.port ||
+  brokerServer.protocol === 'https:' ? '443' : '80';
+const httpsProxy = process.env.https_proxy || process.env.HTTPS_PROXY;
```
adrukh: Prefer CAPS env var over lowercase, like https://github.com/snyk/broker/pull/25/files#diff-6542c5f2ae846b3cb9cf322098613d62R59

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,104 @@
...
+ * Monkey-patch https.request to proxy request to the broker server
+ * according the the env variables `https_proxy` and `no_proxy`:
+ * get the https proxy from the env. proxy requests unless the
+ * broker server is in the no_proxy list.
+ * Start from the "Entry point" below.
+ */
+const url = require('url');
+const tunnel = require('tunnel');
+const {brokerServerUrl} = require('./config');
+const brokerServer = url.parse(brokerServerUrl || '');
```
adrukh: Will this capture the right value regardless of where it came from? I.e., `$BROKER_SERVER_URL` or in the `.env` file?

aviadatsnyk: this is what `lib/config` does https://github.com/snyk/broker/blob/master/lib/config/index.js#L8

lib/patch-https-request-for-proxying.js
aviadatsnyk: fixed.

lib/patch-https-request-for-proxying.js
aviadatsnyk: cleaned. wanted to leave a note for future implementations of `httpsOverHttps` etc.

lib/patch-https-request-for-proxying.js
aviadatsnyk: nice catch! my bad.

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,98 @@
...
+brokerServer.port = brokerServer.port ||
+  brokerServer.protocol === 'https:' ? '443' : '80';
+const httpsProxy = process.env.HTTPS_PROXY || process.env.https_proxy;
```
gjvis: I think we should read this from `require('./config')` – keep env access to a single place

gjvis: (and allow people to use `.env` predictably).

I know that config also modifies process.env, but I think relying on that side-effect is a step too far.

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,98 @@
...
+    return hostnameMatched;
+  });
+}
+
+function shouldProxy(uri) {
+  // Decide the proper request proxy to use based on the request URI object and the
+  // environmental variables (NO_PROXY, HTTP_PROXY, etc.)
+  // respect NO_PROXY environment variables (see: http://lynx.isc.org/current/breakout/lynx_help/keystrokes/environments.html)
+
+  const noProxy = process.env.NO_PROXY || process.env.no_proxy || '';
```
gjvis: likewise here, lets read this via `require('./config')`

lib/patch-https-request-for-proxying.js
aviadatsnyk: fixed.

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  // actual monkey patching: BEWARE!
+  // we're only patching HTTPS requests to the broker server
+  const https = require('https');
+  const oldhttpsreq = https.request;
+  https.request = function (options, callback) {
+    if (options.host === brokerServer.host &&
+        // use == not === to match string and int due to options diffs
+        (!options.port || options.port == brokerServer.port)) {
```
adrukh: Still a bit tricky - if request has no port, I assume it is going to go to port 80? NOT SURE, how can a request have no port?
But in this case, if `brokerServer.port !== 80`, we should not proxy.

gjvis: maybe we're being too clever? we should just proxy everything not in `NO_PROXY`, and not worry about impacting things other than the broker-server? that'd definitely be more predictable

gjvis: Come to think of it, I vote we do that – just proxy everything (apart from `NO_PROXY`) if a proxy is specified. There's way less chance for edge cases we haven't thought of, and its super easy to reason about for users. If it causes problems, then things can still be added to `NO_PROXY` to resolve them.

lib/patch-https-request-for-proxying.js
```diff
@@ -0,0 +1,104 @@
...
+      options.agent = tunnelingAgent;
+    }
+    return oldhttpsreq.call(null, options, callback);
+  };
+}
+
+module.exports = {shouldProxy};   // for testing
```
gjvis: If its worth testing in isolation, then maybe its worthy of its own file?

lib/patch-https-request-for-proxying.js
aviadatsnyk: the original version only looked at `brokerServer.host`. does that sound better?

gjvis: up to you :+1:

PR #26, opened 2017-04-10
fix: ignore case of client broker id

adrukh:
- [x] Ready for review
- [ ] Reviewed by @

#### What does this PR do?

Server is now lower-casing the client identifying token.

Snyk requests made to the broker server always carry a lower-case uuid token, and using an upper-case token on the client config will create a mismatch on the server.


gjvis:
This seems like unexpected behaviour – tokens should be case sensitive

gjvis:
Gah, accidentally closed.

adrukh:
Closing to handle as a breaking change (client identification with server is forcing lowercase on token)

PR #43, opened 2017-06-09
fix: dedicated connection-status endpoint

gjvis:
- [x] Ready for review

#### What does this PR do?

Move the check of client connection-status from `GET /broker/:broker-token` to
`GET /connection-statis/:broker-token`, to avoid interfering with actual
brokered requests.


gjvis:
Hmm, perhaps we should call this out as a breaking change?

PR #45, opened 2017-06-18
fix: allow fetching package-lock.json to support npm5 repos

adrukh:
We want to be able to fetch `package-lock.json` to detect an npm5 project. This change adds a default filter to allow fetching these files both from GitHub and Bitbucket.

In order to make use of this change, either generate a new `accept.json` filter by running `broker init [github|bitbucket]`, or apply this patch to your existing `accept.json` file.

### DIFF
--- a/client-templates/bitbucket-server/accept.json.sample
+++ b/client-templates/bitbucket-server/accept.json.sample
@@ -22,6 +22,12 @@
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
+      "path": "/projects/:project/repos/:repo/browse*/package-lock.json",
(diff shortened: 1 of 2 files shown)


darscan:
I think this is missing two entries in `client-templates/github/accept.json.sample` in the topmost section (`public`). E.g.

```
          {
            "path": "commits.*.added.*",
            "value": "package-lock.json"
          },
          {
            "path": "commits.*.modified.*",
            "value": "package-lock.json"
          },
```

ALSO, we're missing entries for `pom.xml` in the same area.

adrukh:
Indeed!

I'll add `package-lock.json` in this commit. We don't want to know about changes to this file **yet**, but it will be needed in the future.

I'll open separate commits for `pom.xml`, `yarn.lock` etc.

adrukh:
Reopening as a `feat`

PR #65, opened 2017-07-24
feat: add support for gitlab

AH7:
- [x] Ready for review

#### What does this PR do?
Add support for Gitlab servers.

### DIFF
--- a/client-templates/gitlab/.env.sample
+++ b/client-templates/gitlab/.env.sample
@@ -1,0 +1,19 @@
+# your unique broker identifier
+BROKER_TOKEN=
+
+# your personal token to your Gitlab server account
+GITLAB_TOKEN=
+
+# the host where your Gitlab Server is running, excluding scheme.
+# i.e. gitlab.yourdomain.com
+GITLAB=
+
+# the url of your broker client (including scheme and port)
+# BROKER_CLIENT_URL=
+
+# The URL of the Snyk broker server
+BROKER_SERVER_URL=https://broker.snyk.io
+
+# the fine detail accept rules that allow Snyk to make API requests to your
+# gitlab instance
+ACCEPT=accept.json
--- a/client-templates/gitlab/accept.json.sample
+++ b/client-templates/gitlab/accept.json.sample
@@ -1,0 +1,31 @@
+{
+  "private":
+  [
+    {
+      "//": "list the user's projects",
+      "method": "GET",
+      "path": "/api/v3/projects",
+      "origin": "https://${GITLAB}"
+    },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/api/v3/projects/:project/repository/files",
+      "origin": "https://${GITLAB}",
+      "validParams": {
+        "file_path": [
+          "package.json",
+          "yarn.lock",
+          "package-lock.json",
+          "Gemfile",
+          "Gemfile.lock",
+          "pom.xml",
+          "requirements.txt",
+          "build.gradle",
+          "build.sbt",
+          ".snyk"
+        ]
+      }
+    }
+  ]
+}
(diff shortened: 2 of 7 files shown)


client-templates/gitlab/.env.sample
```diff
@@ -0,0 +1,23 @@
...
+
+# the url of your broker client (including scheme and port)
+# BROKER_CLIENT_URL=
+
+# The URL of the Snyk broker server
+BROKER_SERVER_URL=https://broker.snyk.io
+
+# the fine detail accept rules that allow Snyk to make API requests to your
+# bitbucket server instance
+ACCEPT=accept.json
```
adrukh: replace instances of 'bitbucket'?

client-templates/gitlab/accept.json.sample
```diff
@@ -0,0 +1,35 @@
...
+        },
+        {
+          "param": "file_path",
+          "value": ".snyk"
+        },
+        {
+          "param": "ref",
+          "value": "master"
+        }
+      ]
```
adrukh: I suggest a different approach, where each param has a list of possible values for filtering, like so:
```
"validParams": {
  "file_path": ["package.json", "Gemfile.lock", ".snyk"]
}
```

This way you can later filter by possible values in code, see next comment.

client-templates/gitlab/accept.json.sample
```diff
@@ -0,0 +1,35 @@
...
+          "value": "Gemfile.lock"
+        },
+        {
+          "param": "file_path",
+          "value": ".snyk"
+        },
+        {
+          "param": "ref",
+          "value": "master"
+        }
```
adrukh: Why only `master` branch? Any branch should be ok.

lib/filters/index.js
```diff
@@ -75,6 +73,29 @@ module.exports = ruleSource => {
         }
       }
 
+      //hack to support nodejs gitlab client
+      const GITLAB_DUMMY_QUERY =
+        'private_token=6E6ACA18-70B8-4FE3-A602-611209920236';
```
adrukh: Maybe fake with an obvious `private_token=dummy_token`?

lib/filters/index.js
```diff
@@ -75,6 +73,29 @@ module.exports = ruleSource => {
...
+          return validParams.some(({ param, value }) => {
+            return (queryParam == `${param}=${value}`);
+          });
+        });
+
+        if (!isValid) {
+          return false;
+        }
+
+      }
```
adrukh: Instead of this validation (which btw has a problem because if `ref=master` appears, than bad paths are also allowed?), new `"validParams"` object allows this:
```
if (validParams && querystring) {
  return querystring.split('&').all(pair => {
    const [key, val] = pair.split('=');
    if (!validParams[key]) {
      return true;
    }
    return validParams[key].includes(val);
  });
}
```

Not sure if the code above works :)

Plus, it surely doesn't work for nested files, that would be important to fix.

client-templates/gitlab/accept.json.sample
AH7: Cool, looks nicer. thanks :)

client-templates/gitlab/.env.sample
```diff
@@ -0,0 +1,23 @@
+# your unique broker identifier
+BROKER_TOKEN=
+
+# your personal token to your bitbucket server account
+GITLAB_TOKEN=
+
+# the host where your Gitlab Server is running, excluding scheme.
+# for gitlab.yourdomain.com
+# this should be "gitlab.yourdomain.com"
+GITLAB=
```
adrukh: We can probably just use `${GITLAB}` in code, and remove the duplication of `GITLAB_API` from the settings?

lib/filters/index.js
AH7: It will not allow bad paths if ref=master. I'll adopt your proposal. thanks.

PR #89, opened 2017-11-09
feat: rewrite raw endpoints to /contents api for GH

darmalovan:
- [x] Ready for review
- [ ] Follows CONTRIBUTING rules

#### What does this PR do?

Since GH enterprise 2.11 the `/raw` endpoint is not supported.
In order for us to make a gradual change without breaking our users broker clients
I added logic that rewrites /raw to /repos/owner/repo/contents endpoint.
Once all of our GHE users will use the we could change all our system code that touches GH files, or hopefully will do it on the GH agent, and delete this code.

I know that this code makes our filter logic GH aware, and this was not the intention of the broker, but this was the solution for non breaking change.

https://snyksec.atlassian.net/browse/CSUP-4

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ To use the the broker client with a GitHub Enterprise deployment, run `docker pu
 - `GITHUB_TOKEN` - a personal access token with full `repo` and `admin:repo_hook` scopes.
 - `GITHUB` - the hostname of your GitHub Enterprise deployment, such as `your.ghe.domain.com`.
 - `GITHUB_API` - the API endpoint of your GitHub Enterprise deployment. Should be `$GITHUB/api/v3`.
-- `GITHUB_RAW` - the raw file access endpoint of your GitHub Enterprise deployment. Should be `$GITHUB/raw`.
 - `PORT` - the local port at which the broker client accepts connections. Default is 7341.
 - `BROKER_CLIENT_URL` - the full URL of the broker client as it will be accessible by your GitHub Enterprise deployment webhooks, such as `http://my.broker.client:7341`
 
@@ -75,7 +74,6 @@ docker run --restart=always \
            -e GITHUB_TOKEN=secret-github-token \
            -e GITHUB=your.ghe.domain.com \
            -e GITHUB_API=your.ghe.domain.com/api/v3 \
-           -e GITHUB_RAW=your.ghe.domain.com/raw \
            -e PORT=8000 \
            -e BROKER_CLIENT_URL=http://my.broker.client:8000 \
        snyk/broker:github-enterprise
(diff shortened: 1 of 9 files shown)


lib/relay.js
```diff
@@ -55,8 +55,33 @@ function requestHandler(filterRules) {
 function responseHandler(filterRules, config) {
   const filters = Filters(filterRules);
 
+  const supportedFiles = (filterRules || [])
+    .filter(f => f.path && f.path.includes('/contents/'))
+    .map(f => f.path.split('/').pop());
```
adrukh: Let's move this into the `rewriteRawApi` function so that it is clear it is not used outside of its scope.

lib/relay.js
```diff
@@ -55,8 +55,33 @@ function requestHandler(filterRules) {
...
+        headers.accept = 'application/vnd.github.2.11.raw';
+        // switching urls from: /:name/:repo/:branch*/:path
+        // to /repos/:name/:repo/contents/:path?ref=:branch
+        // $1 - repo/owner, $2 - branch, $3 - path
+        url = url
+          .replace(/(\/.+?\/.+?\/)(.+?)(\/.*)/, `/repos$1contents$3?ref=$2`);
+      }
+    }
+
+    rewriteRawApi();
```
adrukh: Call `rewriteRawApi` only if `config.GITHUB` - this way we skip `supportedFiles.find` on every request for non-GH brokers. Remove `config.GITHUB` from line 73.

lib/relay.js
darmalovan: It's a tradeoff of being run all the time, and running it one time.

README.md
```diff
@@ -60,7 +60,7 @@ To use the the broker client with a GitHub Enterprise deployment, run `docker pu
 - `GITHUB_TOKEN` - a personal access token with full `repo` and `admin:repo_hook` scopes.
 - `GITHUB` - the hostname of your GitHub Enterprise deployment, such as `your.ghe.domain.com`.
 - `GITHUB_API` - the API endpoint of your GitHub Enterprise deployment. Should be `$GITHUB/api/v3`.
-- `GITHUB_RAW` - the raw file access endpoint of your GitHub Enterprise deployment. Should be `$GITHUB/raw`.
+- `GITHUB_RAW(*deprecated*, used for versions older than 4.18.5)` - the raw file access endpoint of your GitHub Enterprise deployment. Should be `$GITHUB/raw`.
```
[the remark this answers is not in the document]
darmalovan: Don't you think it should be here for old users? it's just the readme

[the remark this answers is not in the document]
adrukh: Nope I think it will be confusing - we're describing something the user will not need.

lib/relay.js
adrukh: You are right!

PR #100, opened 2018-01-30
feat: add an endpoint for webhooks with ids, get branch data, get user data

aviadatsnyk:
- [X] Ready for review
- [X] Follows CONTRIBUTING rules
- [x] Reviewed by @ … (Snyk internal team)

#### What does this PR do?
Github: Adds endpoints for webhooks with IDs, GET branch and user info.

### DIFF
--- a/client-templates/github/accept.json.sample
+++ b/client-templates/github/accept.json.sample
@@ -68,10 +68,22 @@
             "value": ".snyk"
           }
         ]
-      }
+      },
+    {
+      "//": "used for pushing up webhooks from github",
+      "method": "POST",
+      "path": "/webhook/github/:webhookId"
+    }
   ],
   "private":
   [
+    {
+      "//": "list the user's info, and validate their credentials",
+      "method": "GET",
+      "path": "/user",
+      "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+    },
+
     {
       "//": "list the user's repos",
       "method": "GET",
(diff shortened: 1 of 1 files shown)


client-templates/github/accept.json.sample
```diff
@@ -68,10 +68,22 @@
...
       }
   ],
   "private":
   [
+    {
+      "//": "list the user's info",
+      "method": "GET",
+      "path": "/user",
+      "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+    },
```
adrukh: What do we need the user info for in our agent flow? Creds validation?

client-templates/github/accept.json.sample
```diff
@@ -114,6 +126,13 @@
       "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
     },
 
+    {
+      "//": "used to create a new branch messages",
```
adrukh: ~messages~

adrukh: Do we not do this today via `github-pr`? I wonder how brokered PRs are being opened.

client-templates/github/accept.json.sample
aviadatsnyk: exactly.

client-templates/github/accept.json.sample
aviadatsnyk: we replaced the github api library from https://github.com/pksunkara/octonode to https://github.com/octokit/rest.js

PR #104, opened 2018-02-15
Feat/client metadata

adrukh:
Broker clients will now identify themselves with the broker client by passing their metadata in addition to their token.

The purpose of this change is to improve visibility into live broker clients for better troubleshooting etc.

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -1,15 +1,22 @@
 const socket = require('./socket');
 const relay = require('../relay');
 const logger = require('../log');
+const version = require('../version');
 
 module.exports = ({ port = null, config = {}, filters = {} }) => {
-  logger.info('running');
+  logger.info({ version }, 'running in client mode');
+
+  const identifyingMetadata = {
+    version,
+    filters,
+  };
 
   const io = socket({
     token: config.brokerToken,
     url: config.brokerServerUrl,
     filters: filters.private,
     config,
+    identifyingMetadata,
   });
 
   // start the local webserver to listen for relay requests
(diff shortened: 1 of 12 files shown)


lib/client/socket.js
```diff
@@ -42,8 +42,9 @@ module.exports = ({ url, token, filters, config }) => {
     }
   });
   io.on('open', () => {
-    logger.info({ token, url }, 'identifying');
-    io.send('identify', token);
+    logger.info({ token, url }, 'identifying with broker server');
+    const clientData = { url, token, filters, config };
+    io.send('identify', JSON.stringify(clientData));
```
adrukh: I'm stringifying the client data to parse it on the server side. Should I just send the object as is, and read it on the other side? Didn't check if this is feasible.

gjvis: I'm pretty sure engine-io will handle the serialization for us, so shouldn't need to handle the JSON encoding manually

lib/server/socket.js
```diff
@@ -34,13 +34,26 @@ module.exports = ({ server, filters, config }) => {
...
+      if (!token) {
+        logger.warn({ clientData }, 'client identified without a token');
+        return;
+      }
+
+      logger.info({ clientData, token }, 'client identified');
+
       const clientPool = connections.get(token) || [];
-      clientPool.unshift(socket);
+      clientPool.unshift({ socket, metadata: clientData.metadata });
```
adrukh: connections map now holds objects with socket and metadata

lib/client/index.js
```diff
@@ -1,9 +1,10 @@
 const socket = require('./socket');
 const relay = require('../relay');
 const logger = require('../log');
+const version = require('../../package.json').version || 'local';
```
gjvis: Can we wrap this up into `lib/version.js` please, then reference it appropriately

adrukh: Will do

PR #134, opened 2018-09-07
fix: stronger observability through logging

adrukh:
- [x] Ready for review

#### What does this PR do?

* Adds logging in case of websocket close and error events for debug purposes.
* Removing `debug` dependency, relying solely on `bunyan` for logging.
* Logging *masked* tokens for improved traceability in logs without exposing sensitive data.
* Robust logging when relaying requests (HTTP- and WS-originated)
  * Log on incoming request
  * Log on block by filters
  * Log on relay
  * Log on relay response

### DIFF
--- a/lib/client/socket.js
+++ b/lib/client/socket.js
@@ -38,15 +38,22 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
   io.on('error', ({ type, description }) => {
     if (type === 'TransportError') {
       logger.error({ type, description }, 'Failed to connect to broker server');
+    } else {
+      logger.warn({ type, description }, 'Error on websocket connection');
     }
+
   });
   io.on('open', () => {
-    logger.info({ token, identifyingMetadata },
+    logger.info({ url, token, identifyingMetadata },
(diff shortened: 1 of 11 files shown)


lib/client/socket.js
```diff
@@ -38,15 +38,22 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
...
   io.on('open', () => {
-    logger.info({ token, identifyingMetadata },
+    logger.info({ url, token, identifyingMetadata },
       'identifying with broker server');
     const clientData = { token, metadata: identifyingMetadata };
     io.send('identify', clientData);
   });
 
+  io.on('close', () => {
+    logger.warn({ url, token }, 'server connection closed');
```
joshje: Why a `warn` and not an `info` for this one?

adrukh: From the broker client's perspective, losing the connection to the server is not an 'all good' event. On the other hand, it is also not the client's fault (as far as it knows for now).
Using `warn` log level here to make these logs stand out in case they happen.

PR #135, opened 2018-09-10
feat: Include packages.config in a list of allowed manifest files

lili2311:
- [x] Ready for review

#### What does this PR do?
- Allow a new manifest type `packages.config` for .NET projects

This is part of the planned work around .net & .net core SCM support


lili2311:
I think this is better now?

PR #136, opened 2018-09-10
Feat/healthcheck with conn

adrukh:
- [x] Ready for review

#### What does this PR do?

Broker client healthcheck will now report the websocket connection status. HTTP status code will be 200 for an open websocket to the broker server, 500 otherwise.

Old healthcheck response:
```
{
  ok: true,
  version: '4.35.0'
}
```

New healthcheck response:
```
{
  ok: true | false,
  websocketConnectionOpen: true | false,
  brokerServerUrl: 'https://broker.snyk.io',
  version: '4.36.0'
}
```


lib/client/index.js
```diff
@@ -24,7 +25,18 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
 
   // IMPORTANT: defined before relay (`app.all('/*', ...`)
   app.get(config.brokerHealthcheckPath || '/healthcheck', (req, res) => {
-    return res.status(200).json({ ok: true, version });
+    // io.readyState sets the success of the healthcheck
+    // value of primus.Spark.OPEN means the websocket connection is open
+    const isConnOpen = (io.readyState === primus.Spark.OPEN);
```
joshje: Is it ok to rely on such an internal output of primus? I found the value in code, but not documented:
https://github.com/primus/primus/blob/master/spark.js#L71

adrukh: I huffed and puffed until I found it, indeed somewhat internal. It was discussed over several `primus` issues such as https://github.com/primus/primus/issues/399 and https://github.com/primus/primus/issues/274 (when it was `Primus.OPEN` and not `Primus.Spark.OPEN`).

Since we're locked to `primus@6` for now due to the server & clients setup, I don't see this being changed all of a sudden. Prefer the risk of breakage over an obscure integer value here with some documentation, as it may break just as well.

PR #137, opened 2018-09-11
feat: broker client systemcheck endpoint

adrukh:
- [x] Ready for review

#### What does this PR do?

Adds a `/systemcheck` endpoint to the broker client, which pokes a configured validation URL, asserting network reachability and some credentials correctness of the broker client configuration. Main purpose is to assist in troubleshooting of connectivity / application issues.

Environment variables configuring the systemcheck:
`BROKER_CLIENT_VALIDATION_URL` - the URL to test
`BROKER_CLIENT_VALIDATION_METHOD` - the HTTP method, default `GET`
`BROKER_CLIENT_VALIDATION_TIMEOUT_MS` - the HTTP request timeout, default 5000ms
`BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER` - the `Authorization` header value to be used on the request, default none
`BROKER_CLIENT_VALIDATION_BASIC_AUTH` - the basic auth creds to be base64-encoded and used for the `Authorization` header value, default none

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ To use the the broker client with a Jira deployment, run `docker pull snyk/broke
 - `BROKER_TOKEN` - the snyk broker token, obtained from your Jira integration settings view.
 - `JIRA_USERNAME` - the Jira username.
 - `JIRA_PASSWORD` - the Jira password.
-- `JIRA_BASE_URL` - the URL of your Jira deployment, such as `https://your.jira.domain.com`.
+- `JIRA_HOSTNAME` - the hostname of your Jira deployment, such as `your.jira.domain.com`.
 - `BROKER_CLIENT_URL` - the full URL of the broker client as it will be accessible by your Jira for webhooks, such as `http://my.broker.client:7341`
 - `PORT` - the local port at which the broker client accepts connections. Default is 7341.
 
@@ -201,7 +201,7 @@ docker run --restart=always \
            -e BROKER_TOKEN=secret-broker-token \
            -e JIRA_USERNAME=username \
            -e JIRA_PASSWORD=password \
-           -e JIRA_BASE_URL=https://your.jira.domain.com \
+           -e JIRA_HOSTNAME=your.jira.domain.com \
            -e BROKER_CLIENT_URL=http://my.broker.client:8000 \
            -e PORT=8000 \
        snyk/broker:jira
@@ -217,9 +217,42 @@ FROM snyk/broker:jira
 ENV BROKER_TOKEN        secret-broker-token
 ENV JIRA_USERNAME       username
 ENV JIRA_PASSWORD       password
-ENV JIRA_BASE_URL       https://your.jira.domain.com
+ENV JIRA_HOSTNAME       your.jira.domain.com
 ENV PORT                8000
 ```
+
+### Monitoring
+
+#### Healthcheck
+
+The broker exposes an endpoint at `/healthcheck`, which can be used to monitor the health of the running application. This endpoint responds with status code `200 OK` when the internal request is successful, and returns `{ ok: true }` in the response body.
+
+In the case of the broker client, this endpoint also reports on the status of the broker websocket connection.  If the websocket connection is not open, this endpoint responds with status code `500 Internal Server Error` and `{ ok: false }` in the response body.
+
+To change the location of the healthcheck endpoint, you can specify an alternative path via an environment variable:
+
+```
+ENV BROKER_HEALTHCHECK_PATH /path/to/healthcheck
+```
+
+#### Systemcheck
+
+The broker client exposes an endpoint at `/systemcheck`, which can be used to validate the brokered service (SCM or the like) connectivity and credentials. This endpoint causes the broker client to make a request to a preconfigured URL, and report on the success of the request. The supported configuration is:
+
+* `BROKER_CLIENT_VALIDATION_URL` - the URL to which the request will be made.
+* `BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER` - [optional] the `Authorization` header value of the request. Mutually exclusive with `BROKER_CLIENT_VALIDATION_BASIC_AUTH`.
+* `BROKER_CLIENT_VALIDATION_BASIC_AUTH` - [optional] the basic auth credentials (`username:password`) to be base64 encoded and placed in the `Authorization` header value of the request. Mutually exclusive with `BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER`.
+* `BROKER_CLIENT_VALIDATION_METHOD` - [optional] the HTTP method of the request (default is `GET`).
+* `BROKER_CLIENT_VALIDATION_TIMEOUT_MS` - [optional] the request timeout in milliseconds (default is 5000 ms).
+
+This endpoint responds with status code `200 OK` when the internal request is successful, and returns `{ ok: true }` in the response body. If the internal request fails, this endpoint responds with status code `500 Internal Server Error` and `{ ok: false }` in the response body.
+
+To change the location of the systemcheck endpoint, you can specify an alternative path via an environment variable:
+
+```
+ENV BROKER_SYSTEMCHECK_PATH /path/to/systemcheck
+```
+
 
 ### Advanced Configuration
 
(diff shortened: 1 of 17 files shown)


lib/client/index.js
```diff
@@ -36,7 +39,34 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+    }
+
+    // add the internal validation URL to the response data and check it
+    data.brokerClientValidationUrl = logger.sanitise(config.brokerClientValidationUrl);
+    request(config.brokerClientValidationUrl, { timeout: 5000 }, (error, response) => {
+      if (error) {
+        data.brokerClientValidationUrlSuccess = false;
+        data.error = error.message;
+        data.ok = false;
+        return res.status(500).json(data);
```
joshje: Do we want to know about `isConnOpen` here? I don't see it being added to `data`, so this potentially important info will not be in the response.

lib/client/index.js
```diff
@@ -36,7 +39,34 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+      }
+
+      const goodStatusCode = ((response && response.statusCode) < 400);
+      if (!goodStatusCode) {
+        data.brokerClientValidationUrlSuccess = false;
+        data.brokerClientValidationUrlStatusCode = response && response.statusCode;
+        data.error = 'Internal validation URL status code is not 2xx or 3xx';
+        data.ok = false;
+        return res.status(500).json(data);
```
joshje: Same as above regarding `isConnOpen` being sent back

lib/log.js
```diff
@@ -35,6 +35,14 @@ function sanitise(raw) {
     raw = sanitiseConfigVariable(raw, 'GITLAB_TOKEN');
   }
 
+  if (config.JIRA_USERNAME) {
```
joshje: Oooh, I didn't realise we needed to add these to sanitise here!

How about looking for `Object.keys(config)` containing `USERNAME` or `PASSWORD` to catch them automatically?

lib/client/index.js
adrukh: It appears on the `websocketConnectionOpen` key once `data` is declared.

The code is slightly convoluted as I wanted to stick to the callback-way of calling `request` like in the rest of the codebase. Open to suggestions on how to improve, what do you think would be easier to grok?

lib/log.js
adrukh: Some creds include `TOKEN` and not `USERNAME`. Agree this is too explicit, but didn't want to focus on this part just yet :)

README.md
```diff
@@ -249,7 +249,13 @@ Note that `BROKER_CLIENT_URL` now has the HTTPS scheme.
 
 The broker exposes an endpoint at `/healthcheck`, which can be used to monitor the health of the running application. This endpoint returns `200 OK` status code when the application is healthy, and will return a JSON object containing `ok: true`.
 
-To change the location of this endpoint, you can specify an alternative path via an environment variable:
+In the case of the broker client, this endpoint reports on two more checks:
+* Check on the status of the broker websocket connection. If the connection is not open, the healthcheck fails.
+* [Optional] Make a request to a specified validation URL (`BROKER_CLIENT_VALIDATION_URL`, `BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER` and `BROKER_CLIENT_VALIDATION_METHOD`environment variables). If the request fails (error or 4xx / 5xx status code), the healthcheck fails.
+
+In case of failure, the healtcheck endpoint returns a `500` status code, and the response body elaborates on the failure reasons.
```
adrukh: typo

client-templates/jira/.env.sample
```diff
@@ -7,13 +7,18 @@ JIRA_USERNAME=
 # your personal password to your Jira Server account
 JIRA_PASSWORD=
 
+# your Jira Server hostname, i.e. jira.yourdomain.com
+JIRA_HOSTNAME=
```
adrukh: breaking change for current jira broker users!

dockerfiles/bitbucket-server/Dockerfile
```diff
@@ -59,10 +59,13 @@ ENV ACCEPT accept.json
...
 ENV BROKER_HEALTHCHECK_PATH /healthcheck
 
+# Bitbucket server validation url, checked by broker client healthcheck endpoint
+ENV BROKER_CLIENT_VALIDATION_URL https://$BITBUCKET_USERNAME:$BITBUCKET_PASSWORD@$BITBUCKET/rest/api/1.0/users
+
 
 EXPOSE $PORT
 
-HEALTHCHECK --interval=10s --timeout=1s \
+HEALTHCHECK --interval=60s --timeout=10s \
```
adrukh: Remove docker healthchecks, users should rely on `/healthcheck` instead

lib/client/index.js
```diff
@@ -36,7 +39,47 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+    if (config.brokerClientValidationAuthorizationHeader) {
+      validationRequestHeaders.authorization = config.brokerClientValidationAuthorizationHeader;
+    }
+
+    // make the internal validation request
+    request({
+      url: config.brokerClientValidationUrl,
+      headers: validationRequestHeaders,
+      method: config.brokerClientValidationMethod || 'GET',
+      timeout: 5000,
```
adrukh: make a config value

lib/client/index.js
```diff
@@ -36,7 +39,47 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+    data.brokerClientValidationUrl = logger.sanitise(config.brokerClientValidationUrl);
+    const validationRequestHeaders = {
+      'user-agent': 'Snyk Broker client ' + version,
+    };
+    if (config.brokerClientValidationAuthorizationHeader) {
+      validationRequestHeaders.authorization = config.brokerClientValidationAuthorizationHeader;
+    }
+
+    // make the internal validation request
+    request({
```
adrukh: Make sure `request` follows redirects by default

lib/client/index.js
```diff
@@ -36,7 +39,47 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+      timeout: 5000,
+    }, (error, response) => {
+      if (error) {
+        data.brokerClientValidationUrlSuccess = false;
+        data.error = error.message;
+        data.ok = false;
+        return res.status(500).json(data);
+      }
+
+      const goodStatusCode = ((response && response.statusCode) < 400);
```
adrukh: `/^2/.test(statusCode)`?

lib/client/index.js
joshje: Once I uncollapsed the diff on GH I spotted the declaration! I think the confusing part is actually that `status` refers specifically to the result of `isConnOpen`.

What do you think about calling it `isConnOpenStatus`, or query `isConnOpen` boolean at the point we want to determine the status code?

client-templates/jira/.env.sample
joshje: This won't work in cases where the username has an `@` in it (e.g. an email address). No amount of encoding will help either. This needs to become similar to how we handle the `auth` object as defined in the rules:

```
https://github.com/snyk/broker/blob/master/client-templates/jira/accept.json.sample#L22
```

This is so we can use request's auth handler:
https://github.com/snyk/broker/blob/c0418a3b388e33aa3f2458d3275ae05d1633ce4b/lib/filters/index.js#L137

README.md
```diff
@@ -249,7 +249,13 @@ Note that `BROKER_CLIENT_URL` now has the HTTPS scheme.
```
joshje: If we end up going with `JIRA_HOSTNAME` then make sure to update line 190 above

lib/client/index.js
adrukh: Naming is hard! `isX` is a predicate, so `isConnOpenStatus` doesn't sound proper to me. Maybe `isConnStatusOpen`?

The split flow here is to accommodate the optionality of the validation url, and use `request` in a callback fashion if it is present. Sort of older-school JS... Didn't want to add promise-based `request` lib or similar, or introduce plain `Promise`s just for this flow.

client-templates/jira/.env.sample
adrukh: WOW

I didn't think of this at all!

I did however test this for real with our own jira creds, which included an email address as the username - and it worked! So I'd like to keep it this way, or maybe defer jira validation logic to a later time.

README.md
adrukh: Totally!

adrukh:
I'm going to split the validation into its own endpoint, not to mix healthcheck logic (broker client is responsible for the websocket connection) and systemcheck-like logic (incorrect creds will prevent the broker client from working properly, but it is not the broker client's "fault").

adrukh:
@joshje this is now updated with proper handling of basic auth, thanks for bringing this up!

Please review again. 🙏

PR #138, opened 2018-09-11
feat: allow *.csproj files

lili2311:
- [x] Ready for review

#### What does this PR do?
- Allow a new manifest type `*.csproj` for .NET projects

This is part of the planned work around .net & .net core SCM support

### DIFF
--- a/client-templates/github-com/accept.json.sample
+++ b/client-templates/github-com/accept.json.sample
@@ -90,6 +90,14 @@
           {
             "path": "commits.*.modified.*",
             "value": "packages.config"
+          },
+          {
+            "path": "commits.*.added.*",
+            "value": "*.csproj"
+          },
+          {
+            "path": "commits.*.modified.*",
+            "value": "*.csproj"
           }
         ]
       },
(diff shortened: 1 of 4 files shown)


client-templates/github-com/accept.json.sample
```diff
@@ -90,6 +90,14 @@
...
             "path": "commits.*.modified.*",
             "value": "packages.config"
+          },
+          {
+            "path": "commits.*.added.*",
+            "value": "*.csproj"
+          },
+          {
+            "path": "commits.*.modified.*",
+            "value": "*.csproj"
```
adrukh: I wonder if the filter validation treats `*` as wildcards in the `value` field as well.

client-templates/gitlab/accept.json.sample
```diff
@@ -118,7 +124,8 @@
             "**/build.gradle",
             "**/build.sbt",
             "**/.snyk",
-            "**/packages.config"
+            "**/packages.config",
+            "**/*.csproj"
```
adrukh: Same wildcard question here.

client-templates/github-com/accept.json.sample
lili2311: testing locally, will confirm shortly

lili2311: Verified locally, all events looked good nothing blocked when testing with local webhooks setup.

lili2311:
https://snyksec.atlassian.net/browse/SC-6163

PR #169, opened 2019-05-07
fix: accept filepath with %2F (encoded '/')

yuliabaron:
- [x] Ready for review

#### What does this PR do?
Second part of https://github.com/snyk/broker/pull/168
Some APIs expect encoded slash (%2F) in the filepath. This commit add explicit accept rules for such case

### DIFF
--- a/client-templates/bitbucket-server/accept.json.sample
+++ b/client-templates/bitbucket-server/accept.json.sample
@@ -67,7 +67,29 @@
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
+      "path": "/projects/:project/repos/:repo/browse*%2Fpackage.json",
+      "origin": "https://${BITBUCKET_API}",
+      "auth": {
+        "scheme": "basic",
+        "username": "${BITBUCKET_USERNAME}",
+        "password": "${BITBUCKET_PASSWORD}"
+      }
+    },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
       "path": "/projects/:project/repos/:repo/browse*/package-lock.json",
+      "origin": "https://${BITBUCKET_API}",
+      "auth": {
+        "scheme": "basic",
+        "username": "${BITBUCKET_USERNAME}",
+        "password": "${BITBUCKET_PASSWORD}"
+      }
+    },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/projects/:project/repos/:repo/browse*%2Fpackage-lock.json",
       "origin": "https://${BITBUCKET_API}",
       "auth": {
         "scheme": "basic",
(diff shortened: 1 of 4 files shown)


client-templates/bitbucket-server/accept.json.sample
```diff
@@ -174,6 +284,17 @@
         "password": "${BITBUCKET_PASSWORD}"
       }
     },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/projects/:project/repos/:repo/browse*%2F*.csproj",
```
adrukh: This is one way to allow this, but since `*.csproj` captures `long/path/to/file.csproj`, I think we can simply change `...browse*/*.csproj` to `...browse*.csproj` and capture both the encoded and not encoded cases. Up to you if you wish to redo this.

client-templates/bitbucket-server/accept.json.sample
```diff
@@ -240,6 +416,17 @@
         "password": "${BITBUCKET_PASSWORD}"
       }
     },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/projects/:project/repos/:repo/browse*/vendor%2Fvendor.json",
```
adrukh: The `/` before `vendor` also needs to be encoded here.

client-templates/github-com/accept.json.sample
```diff
@@ -289,217 +289,432 @@
...
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
       "path": "/repos/:name/:repo/contents/:path*/project.json",
       "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
     },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/repos/:name/:repo/contents/:path*/%2Fproject.json",
```
adrukh: the `/` wasn't removed here.

client-templates/gitlab/accept.json.sample
```diff
@@ -163,23 +271,41 @@
...
+            "**%2F*.fsproj",
             "**/project.json",
+            "**%2Fproject.json"
             "**/Gopkg.toml",
+            "**%2FGopkg.toml",
             "**/Gopkg.lock",
-            "**/vendor.json"
+            "**%2FGopkg.lock",
+            "**/vendor.json",
+            "**%2Fvendor.json"
```
adrukh: This is slightly mismatching to the other `vendor.json` rules that insist on `/vendor/vendor.json` paths. This one allows `/foo/bar/vendor.json` as well. This doesn't relate to this specific PR, but we should align on one approach in all rules.

client-templates/github-enterprise/accept.json.sample
```diff
@@ -290,109 +290,215 @@
...
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
       "path": "/repos/:name/:repo/contents/:path*/vendor/vendor.json",
       "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
     },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/repos/:name/:repo/contents/:path*/vendor%2Fvendor.json",
```
adrukh: Encode the `/` before `vendor` too please.

client-templates/gitlab/accept.json.sample
```diff
@@ -51,108 +51,216 @@
...
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
       "path": "/api/v4/projects/:project/repository/files*/vendor.json",
       "origin": "https://${GITLAB}"
     },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/api/v4/projects/:project/repository/files*%2Fvendor.json",
```
adrukh: And here we allow any `vendor.json` file, not just `/vendor/vendor.json`. Outside of the scope of this PR, but needs to be aligned.

client-templates/gitlab/accept.json.sample
yuliabaron: right, will raise that with Boost

client-templates/bitbucket-server/accept.json.sample
yuliabaron: */* enforces a / to be found. I prefer to keep it

PR #170, opened 2019-05-07
fix: add the json verifier to avoid typo errors in json template files

yuliabaron:
- [x] Ready for review

#### What does this PR do?
Added a JSON verifier to make sure accept.json.sample files are valid during the build process

### DIFF
--- a/lintVerifier.sh
+++ b/lintVerifier.sh
@@ -1,0 +1,7 @@
+#/bin/bash
+
+echo "Linting JSON files..."
+for file in $(find ./client-templates -type f -name *.json.sample); do
+    echo "${file}"
+    node -r "fs" -e "JSON.parse(fs.readFileSync('${file}'))"
+done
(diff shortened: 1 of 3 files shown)


lintVerifier.sh
```diff
@@ -0,0 +1,7 @@
+#/bin/bash
+
+echo "Linting JSON files..."
+for file in $(find ./client-templates/ -mindepth 1 -maxdepth 2 -type f -name *.json.sample); do
```
adrukh: Maybe no `-mindepth` and `-maxdepth`, in case we do something different in this folder in the future?

package.json
```diff
@@ -13,7 +13,7 @@
     "start": "node lib",
     "dev": "node lib | ./node_modules/.bin/bunyan",
     "test": "npm run lint && CI=1 tap -R spec test/**/*.test.js --timeout 60",
-    "lint": "eslint -c .eslintrc cli lib test",
+    "lint": "eslint -c .eslintrc cli lib test; ./lintVerifier.sh",
```
adrukh: replace `;` with `&&`
Without it, `eslint` can fail but the process will succeed because the verifier is ok.

PR #175, opened 2019-06-25
fix: make response log less noisy

shaimendel:
- [ ] Ready for review

#### What does this PR do?
Reduce the log level of the `sending response back to websocket connection` message to debug in case the response is 2XX

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -52,7 +52,14 @@ function requestHandler(filterRules) {
       }, ioResponse => {
         logContext.ioStatus = ioResponse.status;
         logContext.ioHeaders = ioResponse.headers;
-        logger.info(logContext, 'sending response back to HTTP connection');
+
+        const logMsg = 'sending response back to HTTP connection';
+        if (ioResponse.status <= 200) {
+          logger.debug(logContext, logMsg);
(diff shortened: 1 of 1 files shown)


lib/relay.js
```diff
@@ -204,7 +204,10 @@ function responseHandler(filterRules, config) {
         const status = (response && response.statusCode) || 500;
         logContext.httpStatus = status;
         logContext.httpHeaders = response.headers;
-        logger.info(logContext, 'sending response back to websocket connection');
+
+        const logFunc = (status >= 400) ? logger.info : logger.debug;
```
adrukh: Please reverse this to:
`const logFunc = (status <= 200) ? logger.debug : logger.info;`

This will support the unlikely case where `status` is `NaN`, and actually log this case at `info` level.
This is because JS is ✨ , and `NaN <= 200` is `false`.

adrukh: Also, I think you need the same treatment for the `sending response back to HTTP connection` which is elsewhere.

adrukh:
Seems like the logger func is missing its ‘this’ object?

adrukh:
Just make this into a nice if statement

PR #180, opened 2019-07-16
feat: improve systemcheck 401 error messages

carlos-snyk:
- [x] Ready for review

#### What does this PR do?

Context on systemcheck: https://github.com/snyk/broker/pull/137
Currently, when systemcheck fails with a 401 (Unauthorized) the error returned is not very helpful.
This PR adds a helpful error message.

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -41,7 +41,7 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
   });
 
   app.get(config.brokerSystemcheckPath || '/systemcheck', (req, res) => {
-    // Systemcheck is the broker client's ability to assert the network 
+    // Systemcheck is the broker client's ability to assert the network
     // reachability and some correctness of credentials for the service
     // being proxied by the broker client.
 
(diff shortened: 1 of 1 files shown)


lib/client/index.js
```diff
@@ -89,9 +89,19 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+      const responseStatusCode = response && response.statusCode;
+      data.brokerClientValidationUrlStatusCode = responseStatusCode;
+
+      if (responseStatusCode === 401) {
+        data.ok = false;
+        data.error = 'The /systemcheck has failed due to invalid credentials. This means the Broker cannot contact your Git.' +
+          ' Please check that the credentials you have used are valid and enable the correct permissions. For more information' +
+          ' about the permissions needed for the Snyk Broker to communicate with your Git, see our documentation.';
+        return res.status(500).json(data);
+      }
```
adrukh: Very nice! A few comments to make it better:
1. The broker proxies more than just `git` services (it can handle `jira`, for example), so perhaps a more generic message?
2. `/systemcheck` is a configurable endpoint, so maybe plug the configured value instead of hardcoding it?
3. I think the most relevant information would actually be in the response body of the 401 response. If we added it to the broker response body with something like `data.brokerClientValidationResponseBody` or similar, that would be best.

carlos-snyk: I've been discussing the error message with @yuliabaron and decided to shorten it to `'The /systemcheck has failed due to invalid credentials.` so that it's correct and more readable.

1. Fixed on the new message
2. Will do
3. That response body can be huge so we thought it might be better to add a `logger.error` with it. What do you think?

adrukh: 1. How about just `Auth error`?
1.5. Consider doing the same for a 403 status code too
2. 👍 
3. Indeed it may, but it carries all the info. Logging it will make it available to users who have access to logs, which is not as easy always. Returning it on our response makes it very convenient.
3.5. Actually, why not return the response body *in any case*? Even if it is a 200 response, showing the response body will help me make sure that the access token used belongs to the user I was intending to see. Sounds like a great troubleshooting improvement in any case.

yuliabaron: 1. Auth error is too short, isn't it?
3. I thought users are always able to read broker client logs... Returning the whole body would break turn the reply format to unreadable ({....}), that's why i thought of adding error. Then on 401 it'll be better already, and we'll log everything anyway so it'll be easy for support to track them. And we'll keep improving the errors over time

carlos-snyk: 1. Added `403` and changed the message to `Failed due to invalid credentials`

3. To keep responses readable, I've added the response body to the logs. As Yulia mentioned, support can track them and we'll keep improving the errors over time.

PR #187, opened 2019-08-19
fix: improve broker error messages

yuliabaron:
- [x] Ready for review

#### What does this PR do?
1. Improves an error message when request is blocked because of the policy
2. Changes the body of the error message from plain text to json format of {message, reason}, where `message` will be the error message sent before and `reason` will be a block reason (when there's no a matching accept rule in accept.json file)

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -33,10 +33,12 @@ function requestHandler(filterRules) {
     logger.debug(logContext, 'received request over HTTP connection');
     filters(req, (error, result) => {
       if (error) {
+        const reason = 'Request does not match any accept rule, blocking HTTP request';
+        error.reason = reason;
         logContext.error = error;
-        logger.info(logContext, 'no rule match, blocking HTTP request');
+        logger.warn(logContext, reason);
         // TODO: respect request headers, block according to content-type
-        return res.status(401).send(error.message);
+        return res.status(401).send({message: error.message, reason});
       }
 
       req.url = result.url;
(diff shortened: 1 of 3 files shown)


lib/relay.js
```diff
@@ -33,10 +33,12 @@ function requestHandler(filterRules) {
...
     filters(req, (error, result) => {
       if (error) {
+        const reason = 'Request does not match any accept rule, blocking HTTP request';
+        error.reason = reason;
         logContext.error = error;
-        logger.info(logContext, 'no rule match, blocking HTTP request');
+        logger.info(logContext, reason);
         // TODO: respect request headers, block according to content-type
-        return res.status(401).send(error.message);
+        return res.status(401).send({message: error.message, reason});
```
adrukh: This changes the response type from plain text to `application/json`, will also affect response headers.

I'd leave this change out for now as it carries more impact to the rest of the system. If you want to improve logging, stick to the logging :)

lib/relay.js
```diff
@@ -33,10 +33,12 @@ function requestHandler(filterRules) {
     logger.debug(logContext, 'received request over HTTP connection');
     filters(req, (error, result) => {
       if (error) {
+        const reason = 'Request does not match any accept rule, blocking HTTP request';
+        error.reason = reason;
         logContext.error = error;
-        logger.info(logContext, 'no rule match, blocking HTTP request');
+        logger.info(logContext, reason);
```
adrukh: Did you want to make this into a `warn` instead of `info`?

lib/relay.js
yuliabaron: This is the request handler, I'd like this info to be sent back to Snyk so we could see this in logz.io, rather than generic `blocked`

PR #191, opened 2019-09-25
feat: added mapping for getting single commit endpoint

AlbertKogan:
- [ ] Ready for review

#### What does this PR do?
Adding new proxy for fetching single commit via [Bitbucket Server API](https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8283038112). The purpose is faster dependency discover process and caching in broker usage cases.

### DIFF
--- a/client-templates/bitbucket-server/accept.json.sample
+++ b/client-templates/bitbucket-server/accept.json.sample
@@ -615,6 +615,17 @@
       }
     },
     {
+      "//": "get a specific commit identified by the commit hash or name of a branch",
+      "method": "GET",
+      "path": "/rest/api/1.0/projects/:project/repos/:repo/commits/:sha",
+      "origin": "https://${BITBUCKET}",
+      "auth": {
+        "scheme": "basic",
+        "username": "${BITBUCKET_USERNAME}",
+        "password": "${BITBUCKET_PASSWORD}"
+      }
+    },
+    {
       "//": "used to create a Code Insights report",
       "method": "PUT",
       "path": "/rest/insights/1.0/projects/:project/repos/:repo/commits/:sha/reports/:report",


client-templates/bitbucket-server/accept.json.sample
```diff
@@ -701,6 +701,17 @@
         "username": "${BITBUCKET_USERNAME}",
         "password": "${BITBUCKET_PASSWORD}"
       }
+    },
+    {
+      "//": "get a specific commit identified by the commit hash or name of a branch",
+      "method": "GET",
+      "path": "/rest/api/1.0/projects/:project/repos/:repo/commits/:sha",
```
carlos-snyk: would prefer if this block was next to similar paths, it helps when reviewing

.gitignore
```diff
@@ -4,3 +4,6 @@
 coverage
 /test/fixtures/certs/tmp
 accept.json
+
+# IDE
+.idea
```
carlos-snyk: You could use a global gitignore for your IDE files: http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

AlbertKogan: Fair enough! Thanks!

client-templates/bitbucket-server/accept.json.sample
AlbertKogan: You mean using `BITBUCKET_API` instead? For some reason didn't work

carlos-snyk: no, I just meant moving it next to similar blocks `/rest/api/1.0/projects/:project/repos/:repo/...`

Personal preference, no need to update if it doesn't sound good 😄

yuliabaron: I agree with @carlos-snyk

AlbertKogan: Got it. Will fix, no problem

PR #193, opened 2019-10-08
feat: Support for binary data streaming

mladkau:
Adds support for streaming of binary data. Using the `data` and `end` events the request response on the client side is switched to streaming and each received `chunk` is send over the websocket. The server forwards each received `chunk` directly to the consuming service.

### DIFF
--- a/lib/client/socket.js
+++ b/lib/client/socket.js
@@ -30,10 +30,12 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
 
   logger.info({ url }, 'broker client is connecting to broker server');
 
-  const response = relay.response(filters, config);
+  const response = relay.response(filters, config, io);
+  const streamingResponse = relay.streamingResponse;
 
   // RS note: this bind doesn't feel right, it feels like a sloppy way of
   // getting the filters into the request function.
+  io.on('chunk', streamingResponse(token));
   io.on('request', response());
   io.on('error', ({ type, description }) => {
     if (type === 'TransportError') {
@@ -53,7 +55,7 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
   io.on('close', () => {
     logger.warn({ url, token }, 'websocket connection to the broker server was closed');
   });
-  
+
   // only required if we're manually opening the connection
   // io.open();
 
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -9,6 +9,7 @@ const tryJSONParse = require('./try-json-parse');
 const logger = require('./log');
 const version = require('./version');
 const stream = require('stream');
+const NodeCache = require('node-cache');
 
 module.exports = {
   request: requestHandler,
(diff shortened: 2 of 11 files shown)


lib/client/socket.js
```diff
@@ -30,10 +30,12 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
 
   logger.info({ url }, 'broker client is connecting to broker server');
 
-  const response = relay.response(filters, config);
+  const response = relay.response(filters, config, io);
+  const streamingResponse = relay.streamingResponse;
 
   // RS note: this bind doesn't feel right, it feels like a sloppy way of
   // getting the filters into the request function.
+  io.on('chunk', streamingResponse(token));
```
mladkau: We have now a new `chunk` message which contains chunks of a data stream.

lib/relay.js
```diff
@@ -8,12 +8,41 @@ const replace = require('./replace-vars');
...
+const stream = require('stream');
 
 module.exports = {
   request: requestHandler,
   response: responseHandler,
+  streamingResponse: streamResponseHandler,
 };
 
+const binaryRequest = [
+  'application/vnd.docker.image.rootfs.diff.tar.gzip',
```
mladkau: For now only very specific `accept` headers will trigger the streaming code.

lib/relay.js
```diff
@@ -8,12 +8,41 @@ const replace = require('./replace-vars');
...
+];
+
+const streams = new Map();
```
mladkau: Maybe there is a better way to bind the server response to the chunk handler. I do think the connection pooling gets in the way and I haven't found a way to remove event listeners from a socker i..e `off(streamResponseHandler)`

For now each response object of a binary streaming request gets a UUID which the streamResponseHandler uses to send the incoming chunks into the right response.

lib/relay.js
```diff
@@ -43,6 +72,29 @@ function requestHandler(filterRules) {
 
       req.url = result.url;
       logContext.ioUrl = result.url;
+
+      // check if this is a streaming request for binary data
+      if (req.headers && binaryRequest.indexOf(req.headers['accept']) !== -1) {
```
mladkau: Identify the streaming request by the accept header.

lib/relay.js
```diff
@@ -43,6 +72,29 @@ function requestHandler(filterRules) {
...
+
+        res.locals.io.send('request', {
+          url: req.url,
+          method: req.method,
+          body: req.body,
+          headers: req.headers,
+          streamingID,
+        });
+
+        return;
```
mladkau: We don't set header or anything else in the response as the whole response is streamed directly from the remote system. All filtering must have happened at this point.

lib/relay.js
```diff
@@ -197,15 +251,31 @@ function responseHandler(filterRules, config) {
...
-      }, (error, response, body) => {
+      };
+
+      // check if this is a streaming request for binary data
+      if (streamingID) {
+        logger.debug(logContext, 'serving binary stream request');
+
+        req.encoding = null; // no encoding for binary data
+
+        request(req).on('data', chunk => {
```
mladkau: As mentioned above - we pipe all chunks of the response directly to the broker server which then forwards everything directly to the consuming service.

lib/relay.js
adrukh: consider `req.get('accept')` instead?

jurglic: Is `req.headers &&` necessary?

lib/relay.js
```diff
@@ -197,37 +250,72 @@ function responseHandler(filterRules, config) {
...
+        }).on('error', error => {
+          logError(logContext, error);
+          io.send('chunk', streamingID, error.message, true, {
+            status : 500
+          });
+        });
+        return;
+      }
+
+      request(req, (error, response, body) => {
```
mladkau: I've tried to merge the two request statements into one but this got quite messy and didn't reduce the code much. I feel that the two mechanisms are distinctive enough to warrant two branches.

test/fixtures/server/filters.json
```diff
@@ -28,10 +28,21 @@
...
+      "method": "GET",
+      "origin": "http://localhost:${originPort}",
     }
 
   ],
   "public": [
+    {
+      "path": "/test-blob/*",
+      "method": "GET",
+      "stream": true
```
mladkau: Changing this flag will fail the tests as the response will be encoded by the non-streaming code.

lib/relay.js
moshikod: I think this is fairly a correct approach to handle this async collection of binary chunks.
Consider adding some mechanism to limit the maximum memory consumption together with an ageing limit to handle the case where chunks streaming was stopped from some reason.

test/utils.js
```diff
@@ -15,6 +15,27 @@ const { app: echoServer, server } = webserver({
   httpsCert: process.env.TEST_CERT, // Optional
 });
 
+echoServer.get(
+  '/test-blob/1',
+  (req, res) => {
+    res.setHeader('test-orig-url', req.originalUrl);
+    res.status(299);
```
moshikod: I there any special reason for using 299?

lib/relay.js
mladkau: I think the aging limit is a fair point and I'll add a TTL to the entries (using node-cache). 

In terms of memory consumption: We create a streamingID for each binary download - we would need a fair amount of downloads before we reach any kind of significant memory limit (the streaming objects are only PassThrough objects and should not consume a lot of memory). I think I would prefer to wait here and see if we run into memory issues before considering this.

PR #196, opened 2019-10-31
feat: Use Broker ID during initial client registration

mladkau:
This PR adds the Broker ID to the initial connection URL when a client connects to the server. This change is required for load-balancing requests to the server.

### DIFF
--- a/lib/client/socket.js
+++ b/lib/client/socket.js
@@ -2,13 +2,6 @@ require('../patch-https-request-for-proxying');
 
 const Primus = require('primus');
 const relay = require('../relay');
-const Socket = Primus.createSocket({
-  transformer: 'engine.io',
-  parser: 'EJSON',
-  plugin: {
-    'emitter': require('primus-emitter'),
-  }
-});
 const logger = require('../log');
 
 module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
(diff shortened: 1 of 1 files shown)


lib/client/socket.js
```diff
@@ -26,6 +19,15 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
     throw error;
   }
 
+  const Socket = Primus.createSocket({
+    transformer: 'engine.io',
+    parser: 'EJSON',
+    plugin: {
+      'emitter': require('primus-emitter'),
+    },
+    pathname : '/primus/'+token,
```
mladkau: In the moment this defaults to just /primus (see: https://github.com/primus/primus/blob/master/index.js#L67) but the server will accept anything under /primus. The token can be picked up by a load balancer to route the request. See: https://github.com/snyk/ops/blob/master/helm/broker-container-server/templates/load-balancing.yaml#L13

JackuB: What are we going to load balance? Isn't it stateless thus current round-robin should be enough? Maybe I'm missing some context for this PR.

mladkau: The broker is not stateless as Agent and Broker Client need to access the same server instance to be able to communicate.

JackuB:
```suggestion
    pathname: `/primus/${token}`,
```

PR #197, opened 2019-11-01
feat: connection status return filters

carlos-snyk:
- [ ] Ready for review

#### What does this PR do?

The connection-status endpoint will now return a list of filters. 

This will help debugging Broker issues caused by the filters (usually outdated filters).

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -169,7 +169,7 @@ function responseHandler(filterRules, config, io) {
 
     filters({ url, method, body, headers }, (error, result) => {
       if (error) {
-        const reason = 'Response does not match any accept rule, blocking websocket request';
+        const reason = 'Request does not match any accept rule, blocking websocket request';
         logContext.error = error;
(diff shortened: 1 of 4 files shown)


lib/relay.js
```diff
@@ -169,7 +169,7 @@ function responseHandler(filterRules, config, io) {
 
     filters({ url, method, body, headers }, (error, result) => {
       if (error) {
-        const reason = 'Response does not match any accept rule, blocking websocket request';
+        const reason = 'Request does not match any accept rule, blocking websocket request';
```
carlos-snyk: this was not a response and confused me 😕

adrukh: Please undo this change, this is indeed a response. The naming is a bit convoluted because the relay logic applies both to the server and the client. A response in this case is the answer being relayed back to the entity that ‘asked the question’, and will be dropped if it isn’t filtered.

PR #204, opened 2019-11-07
chore: lint accept filters on CI

adrukh:
- [x] Ready for review

#### What does this PR do?

Applies linting to sample accept filters on CI. Will help us release proper accept filters!

### DIFF
--- a/lintVerifier.sh
+++ b/lintVerifier.sh
@@ -1,4 +1,6 @@
-#/bin/bash
+#! /usr/bin/env bash
+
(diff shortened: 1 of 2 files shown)


lintVerifier.sh
```diff
@@ -1,4 +1,6 @@
-#/bin/bash
+#! /usr/bin/env bash
+
+set -eo pipefail
```
adrukh: This makes the script bail out on the first command that has a return code != 0

PR #205, opened 2019-11-08
fix: Adding client templates for Artifactory Container Registry

mladkau:
This PR adds the client templates for Artifactory Container Registry

### DIFF
--- a/client-templates/artifactory-cr/accept.json.sample
+++ b/client-templates/artifactory-cr/accept.json.sample
@@ -1,0 +1,30 @@
+{
+  "public":
+  [
+    {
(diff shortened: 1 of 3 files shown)


client-templates/artifactory-cr/accept.json.sample
```diff
@@ -0,0 +1,30 @@
...
+    {
+      "//": "send any type of request to our connected server",
+      "method": "any",
+      "path": "/*"
+    }
+  ],
+  "private":
+  [
+    {
+      "//": "Get the about page",
```
adrukh: @mladkau maybe we'd want to be more precise here :)

PR #207, opened 2019-11-12
feat: allow gradle.properties file

lili2311:
- [x] Ready for review

#### What does this PR do?
Allow us to grab the gradle.properties file to help with gradle
project dependency generation.

We are starting to roll out this file to be picked up on every test run if present. It is currently rolled out to 10%
Adding broker support as well

### DIFF
--- a/client-templates/gitlab/accept.json.sample
+++ b/client-templates/gitlab/accept.json.sample
@@ -150,6 +150,18 @@
     {
       "//": "used to determine the full dependency tree",
       "method": "GET",
+      "path": "/api/v4/projects/:project/repository/files*/gradle.properties",
+      "origin": "https://${GITLAB}"
+    },
+    {
+      "//": "used to determine the full dependency tree",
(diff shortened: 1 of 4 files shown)


client-templates/gitlab/accept.json.sample
```diff
@@ -144,7 +144,19 @@
...
+        {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/api/v4/projects/:project/repository/files*/gradle.properties",
+      "origin": "https://${GITLAB}"
+    },
+    {
+      "//": "used to determine the full dependency tree",
+      "method": "GET",
+      "path": "/api/v4/projects/:project/repository/files*%2Fgradle.properties",
```
yuliabaron: first and third look the same?

lili2311: fixed, accidentally updated a wrong one

PR #208, opened 2019-11-14
fix: node 8->12, debian 2017->2019

FauxFaux:
#### What does this PR do?

Jump to the current node LTS (12), and also upgrade the underlying Debian image for our dockerfiles from Debian 2017 (the default) to Debian Jun 2019 ("buster") (the new stable version).

The dockerfiles build, but print a couple of warnings:
 1) Can't build dtrace, because `python` isn't installed. This isn't new, and is probably fine.
 2) `Unsupported engine for got@5.7.1: wanted: {"node":">=0.10.0 <7"}`

.. which also would have been broken before.


FauxFaux:
`got` is from `clite`, a CLI framework, which has been abandoned for four years.

FauxFaux:
It starts, and handles http requests.

Do we have a test plan, e.g. against JIRA?

yuliabaron:
Not really. I'd do the standard flow with this broker (import, Open Fix PR, make sure Test on PR works). If all this is fine - we can merge

adrukh:
We better get an `.nvmrc` here too?

PR #209, opened 2019-11-18
Fix/bump deps

adrukh:
- [x] Ready for review

#### What does this PR do?

* Updates dependencies according to defined semver ranges
* Removes lockfile, since the broker is installed as a library, and its lockfile isn't used with `npm i`.


.gitignore
```diff
@@ -4,3 +4,4 @@
 coverage
 /test/fixtures/certs/tmp
 accept.json
+package-lock.json
```
yuliabaron: why do we ignore the lock file?

adrukh: > One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

From https://docs.npmjs.com/files/package-lock.json

adrukh: So when someone runs `npm i -g snyk-broker`, they get the `package.json` only, and fetch deps according to it (not the lockfile).

yuliabaron: But don't we ourselves require lock file to be pushed so we can monitor it? and if someone adds a new dependency - don't we want to have the exact version used?

adrukh: As discussed, lockfile doesn't get published to npmjs, so keeping it in the repo is actually misleading as to the deps that will be installed with this package.

PR #211, opened 2019-11-21
Use CA_CERT for /systemcheck endpoint

JackuB:
`CA_CERT` envvar was ignored for `/systemcheck` endpoint, which resulted in false negatives when checking whether self-signed certificates are setup correctly

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -75,6 +75,9 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
       method: brokerClientValidationMethod,
       timeout: brokerClientValidationTimeoutMs,
       json: true,
+      agentOptions: {
+        ca: config.caCert, // Optional CA cert
+      },
     }, (error, response) => {
       // test logic requires to surface internal data
       // which is best not exposed in production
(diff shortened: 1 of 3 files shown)


lib/client/index.js
```diff
@@ -75,6 +75,9 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
       method: brokerClientValidationMethod,
       timeout: brokerClientValidationTimeoutMs,
       json: true,
+      agentOptions: {
```
yuliabaron: waaaaaaat??? that's all?!

test/functional/client-ca.test.js
```diff
@@ -62,7 +63,9 @@ test('correctly use supplied CA cert on client for connections', t => {
 
         // Specify CA file
         process.env.CA_CERT = '../certs/ca/my-root-ca.crt.pem';
-        client = app.main({ port: port() });
+        process.env.BROKER_CLIENT_VALIDATION_URL = `https://localhost:${echoServerPort}/test`;
```
yuliabaron: how did the test pass before your fix?

JackuB: That URL is used only the for /systemcheck endpoint, which was never used in tests. With correct CA setup, Broker works, but if you manually invoke /systemcheck, you get an error.

PR #214, opened 2020-01-10
fix: log sanitization of sensitive info

jurglic:
We have found out that some tokens can be unintentionally logged by broker client through recently added httpUrl and ioUrl log keys. This regression was introduced in 5a8213afc.
    
Fix this, and also add a test for keys sanitization.

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -64,6 +64,8 @@ const log = bunyan.createLogger({
     token: sanitise,
     result: sanitise,
     url: sanitise,
+    httpUrl: sanitise,
+    ioUrl: sanitise,
     headers: sanitiseHeaders,
   },
 });
--- a/test/unit/log.test.js
+++ b/test/unit/log.test.js
@@ -1,0 +1,57 @@
+const test = require('tap-only');
+const stream = require('stream');
+
+test('log sanitization of sensitive information', t => {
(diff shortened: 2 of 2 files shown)


lib/log.js
```diff
@@ -39,7 +39,7 @@ function sanitise(raw) {
     raw = sanitiseConfigVariable(raw, 'JIRA_USERNAME');
   }
 
-  if (config.JIRA_USERNAME) {
```
jjmschofield: are sure this is the right thing? I'm not familiar with the config for the broker though (eg is USERNAME actually a misnomer here?)

test/unit/log.test.js
```diff
@@ -0,0 +1,39 @@
...
+  const bbPass = process.env.BITBUCKET_PASSWORD = 'BB_PASS';
+  const jiraUser = process.env.JIRA_USERNAME = 'JRA_USER';
+  const jiraPass = process.env.JIRA_PASSWORD = 'JRA_PASS';
+
+  const sensitiveInfo =
+    [brokerTk, githubTk, gitlabTk, bbUser, bbPass, jiraUser, jiraPass].join();
+
+  // prepare logger and capture its output
+  const logger = require('../../lib/log');
+  const stdoutWriteSpy = sinon.spy(process.stdout, 'write');
```
jjmschofield: this spy should probably be restored in test teardown?

PR #226, opened 2020-03-17
fix: discard any information in url after a fragment identifier

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?
Currently information after a fragment identifier in the url provided to the broker client is not sanitised and could theoretically be exploited (only by a malicious Snyk Broker Server) with a path that matches a filter because it has a manifest file after the fragment.

This sanitises the URL by removing any information found after a fragment identifier (`#`) in the provided path.

See the new tests in `test/unit/filters.test.js`

A request to broker client providing the following path should fail:
`/repos/owner/repo/contents/sensitive-file.js#/package.json`

### DIFF
--- a/lib/filters/index.js
+++ b/lib/filters/index.js
@@ -64,8 +64,12 @@ module.exports = ruleSource => {
       if (req.method.toLowerCase() !== method && method !== 'any') {
         return false;
       }
+
+      // Discard any fragments before further processing
+      const mainURI = req.url.split('#')[0];
+
       // query params might contain additional "?"s, only split on the 1st one
-      const parts = req.url.split('?');
+      const parts = mainURI.split('?');
       let [url, querystring] = [parts[0], parts.slice(1).join('?')];
       const res = regexp.exec(url);
       if (!res) {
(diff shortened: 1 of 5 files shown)


lib/filters/index.js
```diff
@@ -64,8 +64,12 @@ module.exports = ruleSource => {
       if (req.method.toLowerCase() !== method && method !== 'any') {
         return false;
       }
+
+      // Discard any fragments before further processing
+      const mainURI = req.url.split('#')[0];
```
carlos-snyk: Could use the `url` module:
```
> const url = require('url')
> url.parse('/hello/world?some=test').path
'/hello/world?some=test'
> url.parse('/hello/world#hello?some=test').path
'/hello/world'
```

lib/filters/index.js
```diff
@@ -140,11 +144,12 @@ module.exports = ruleSource => {
     };
   });
 
-  return (url, callback) => {
+  return (payload, callback) => {
```
yuliabaron: isn't it the `url` from above?

iamtmrobinson: Confusingly, no. It's a payload object that has a `url` key inside it.

lib/filters/index.js
iamtmrobinson: Yep, I think it's ideal to use that library (e.g. in case there are other edge cases), but I wanted to keep this PR as light-touch as possible to close the known gap and come back to that after.

lib/filters/index.js
carlos-snyk: It's hard to follow but the `tests` functions take a `req`, which is a socket payload: 
https://github.com/snyk/broker/blob/907a63baa2e183f612ab0824beea39b1876b84d4/lib/relay.js#L99-L105

Would be good to refactor these variable names in a separate PR

PR #227, opened 2020-03-17
feat: add TS (allow JS for now)

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?
Adds Typescript to the Broker. JS is allowed for now and nothing has been updated to actually _use_ TS, there's just some basic linting updates.

This has been tested by `npm pack`ing a local copy and building a docker image from scratch, which ran just fine.

Ensure broker runs as usual.


package.json
```diff
@@ -10,10 +13,13 @@
...
-    "start": "node lib",
-    "dev": "node lib | ./node_modules/.bin/bunyan",
-    "test": "npm run lint && CI=1 tap -R spec test/**/*.test.js --timeout 60",
-    "lint": "eslint -c .eslintrc cli lib test && ./lintVerifier.sh",
+    "build": "tsc",
+    "start": "node dist",
+    "dev": "tsc-watch --project tsconfig.json --onSuccess 'node .' | ./node_modules/.bin/bunyan",
+    "test": "npm run lint && CI=1 tap -R spec test/**/*.test.[tj]s --timeout 60",
+    "lint": "npm run lint:js && npm run lint:ts",
+    "lint:ts": "tslint --project tsconfig.json --format stylish",
```
carlos-snyk: `eslint` now does TS, as seen in the `.eslintrc.js` you added, shouldn't need `tslint`.

`tslint` is deprecated: https://github.com/palantir/tslint/issues/4534

lib/patch-https-request-for-proxying.js
```diff
@@ -30,7 +30,7 @@ function parseNoProxyZone(zone) {
   return {hostname: zoneHost, port: zonePort, hasPort: hasPort};
 }
 
-function uriInNoProxy(uri, noProxy) {
+function uriInNoProxy(uri) {
   const port = uri.port || (uri.protocol === 'https:' ? '443' : '80');
   const hostname = formatHostname(uri.hostname);
   const noProxyList = noProxy.split(',');
```
yuliabaron: where this `noProxy` is coming from now?

test/unit/filters.test.js
```diff
@@ -161,7 +161,7 @@ test('filter on body', t => {
       method: 'POST',
       body: jsonBuffer({
         query: `{
-        repositoryOwner(login: "search: "{\"username\": {\"$regex\": \"sue\"}, \"email\": {\"$regex\": \"sue\"}}"") {
+        repositoryOwner(login: "search: "{"username": {"$regex": "sue"}, "email": {"$regex": "sue"}}"") {
```
yuliabaron: wondering why `"` were escaped before...

lib/patch-https-request-for-proxying.js
iamtmrobinson: It's imported at the top of the file, and the call site for this function was just passing it through. Now this function just uses the globally available one rather than having it passed in.

package.json
iamtmrobinson: Thanks for confirming, this one confused me abit. Think the Notion doc might be a little out of date so will update that too.

test/unit/filters.test.js
iamtmrobinson: Think you're right, important for the test. Have brought it back and imported `gql-tag` to satisfy the linting.

PR #228, opened 2020-03-18
feat: reject paths with path traversal

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?
The GitHub API will follow dot-dot-slash (`../`) sequences and there is currently no sanitisation of this in the broker.
Whilst none of the existing examples suggest to do this, if a client was to whitelist a folder (e.g. `/path/to/docs/*`) then path traversal could be used to access a sensitive file e.g.
`/path/to/docs/../../path/to/sensitive/file.js`.

This changes that so that if a [normalized path](https://nodejs.org/api/path.html#path_path_normalize_path) does not match the original path then the request is rejected.

### DIFF
--- a/lib/filters/index.js
+++ b/lib/filters/index.js
@@ -1,6 +1,7 @@
 const minimatch = require('minimatch');
 const pathRegexp = require('path-to-regexp');
 const qs = require('qs');
+const path = require('path');
 const undefsafe = require('undefsafe');
 const replace = require('../replace-vars');
 const authHeader = require('../auth-header');
@@ -32,7 +33,7 @@ module.exports = ruleSource => {
   // array of entries with
   const tests = rules.map(entry => {
     const keys = [];
-    let { method, origin, path, valid, stream } = entry;
+    let { method, origin, path: entryPath, valid, stream } = entry;
     method = (method || 'get').toLowerCase();
     valid = valid || [];
 
@@ -44,7 +45,7 @@ module.exports = ruleSource => {
     const fromConfig = {};
 
     // slightly bespoke version of replace-vars.js
-    path = (path || '').replace(/(\${.*?})/g, (_, match) => {
+    entryPath = (entryPath || '').replace(/(\${.*?})/g, (_, match) => {
       const key = match.slice(2, -1); // ditch the wrappers
       fromConfig[key] = config[key] || '';
       return ':' + key;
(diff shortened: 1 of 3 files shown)


lib/filters/index.js
```diff
@@ -1,6 +1,7 @@
 const minimatch = require('minimatch');
 const pathRegexp = require('path-to-regexp');
 const qs = require('qs');
+const path = require('path');
```
darscan: Could get a smaller diff by just importing this differently, e.g. `const pathLib = require('path');`

test/fixtures/accept/github.json
```diff
@@ -8,6 +8,12 @@
       "method": "GET",
       "path": "/repos/:name/:repo/contents/:path*/package.json",
       "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+    },
+    {
```
yuliabaron: so why do you need to add a rule to whitelist a folder after all?

lib/filters/index.js
iamtmrobinson: I went with this option because I prefer to call packages by their 'real' name as I just find that easier to grok.

test/fixtures/accept/github.json
iamtmrobinson: As in what's the use case for whitelisting directories? I'm not sure 😂 but at least if someone _did_ then the directory can't be traversed 😬

wyc-thg: Someone might have all their dependency files in a "safe" folder or it might be part of a meta-project in a monorepo.

lib/filters/index.js
darscan: > I went with this option because I prefer to call packages by their 'real' name

That means you let other people decide your variable names :D

> I just find that easier to grok

Doesn't your editor `grok` stuff for you? Anyway, each to their own :D

test/fixtures/accept/github.json
yuliabaron: we usually do not add rules which we do not use. that's why adding only specific files' rules. customers then can edit this file per what they need to.

iamtmrobinson: Just to clarify this is only a rule in a text fixture to prove that the exploit is not possible. It's not modifying the `accept.json` we provide as a template for others to extend

yuliabaron: oh, I missed the file name. NP then

PR #234, opened 2020-04-02
feat: remove Accept-Encoding header to not corrupt gzip encoded requests

dkontorovskyy:
#### What does this PR do?
Remove Accept-Encoding header to not corrupt gzip encoded requests

While working with @darscan on this we found out that broker was never passing back into the client compressed content. So broker-client was always decompressing it and sending back the decoded content, but leaving in headers 'Content-Encoding': 'gzip,deflate'. This was making npm cli to freak out as it couldn't decode already decoded content.

This PR removes 'Accept-Encoding' header which means that content will always be plain for now

Renaming and adding test from that PR
https://github.com/snyk/broker/pull/221

### DIFF
--- a/test/functional/server-client.test.js
+++ b/test/functional/server-client.test.js
@@ -44,7 +44,7 @@ test('proxy requests originating from behind the broker server', t => {
   server.io.on('connection', socket => {
     socket.on('identify', clientData => {
       const token = clientData.token;
-      t.plan(24);
+      t.plan(26);
 
       t.test('identification', t => {
         const filters = require(`${clientRootPath}/${ACCEPT}`);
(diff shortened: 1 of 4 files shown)


test/functional/server-client.test.js
```diff
@@ -289,6 +289,29 @@ test('proxy requests originating from behind the broker server', t => {
         });
       });
 
+      t.test('ignores accept-encoding', t => {
```
darscan, line 292: ignores accept-encoding (gzip)

test/functional/server-client.test.js
```diff
@@ -289,6 +289,29 @@ test('proxy requests originating from behind the broker server', t => {
...
+        const msg = 'hello-'.repeat(200);
+        const url = `http://localhost:${serverPort}/broker/${token}/echo-param/${msg}`;
+        request({ url, method: 'get', gzip: true }, (err, res) => {
+          t.equal(res.statusCode, 200, '200 statusCode');
+          t.equal(res.body, msg, 'body');
+          t.end();
+        });
+      });
+
+      t.test('ignores accept-encoding', t => {
```
darscan, line 303: ignores accept-encoding (deflate)

darscan:
Please split the "why" part from the commit title into the commit body
```
feat: Remove Accept-Encoding header

- Compression is not yet supported by the broker
- So, explicitly drop support for compression
```

test/functional/server-client.test.js
```diff
@@ -289,6 +289,29 @@ test('proxy requests originating from behind the broker server', t => {
         });
       });
 
+      t.test('ignores accept-encoding (gzip)', t => {
+        // Create "big" param so it would trigger sending compressed response
+        const msg = 'hello-'.repeat(200);
```
iamtmrobinson, line 294: Nitpick: I'd pull this out to a variable that explains the same thing e.g. `const paramRequiringCompression = 'hello-'.repeat(200);` to avoid duplication and comments that can get out of sync with code.

PR #241, opened 2020-04-16
feat: improve auditability of broker

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?

To improve auditability of the Broker client, this raises the log level of HTTP200 responses to INFO so they are included in the logs by default.

It also adds a new config parameter which allows the SCM response body to be logged by the client if required.

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -121,7 +121,7 @@ function requestHandler(filterRules) {
         method: req.method,
         body: req.body,
         headers: req.headers,
-        streamingID : '',
+        streamingID: '',
       }, ioResponse => {
         logContext.ioStatus = ioResponse.status;
         logContext.ioHeaders = ioResponse.headers;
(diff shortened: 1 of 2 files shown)


lib/relay.js
```diff
@@ -316,23 +316,19 @@ function responseHandler(filterRules, config, io) {
...
-  const logMsg = 'sending response back to websocket connection';
-  if (status <= 200) {
-    logger.debug(logContext, logMsg);
-  } else {
-    logger.info(logContext, logMsg);
-  }
+  logContext.httpHeaders = response.headers;
+  logContext.httpBody = config && config.LOG_ENABLE_BODY === 'true' ? response.body : null;
+
+  logger.info(logContext, 'sending response back to websocket connection');
```
yuliabaron, line 331: is this only used in broker client?

iamtmrobinson: Yep, the server part (above) just calls the logger directly.

PR #242, opened 2020-04-17
feat: add request header filtering

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?

Adds the ability to filter by request headers using `accept.json`. The filtering is strict so, rather than just matching values, if a header is defined in the whitelist and the request does not include the expected header key it will fail.

Make a brokered request to the GitHub Commits API using the `application/vnd.github.cryptographer-preview` header and it will get blocked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,33 @@ docker run --restart=always \
        snyk/broker:github-com
 ```
 
+#### Types of Filtering
+
+##### By Header
+
+Add a validation block with the following key/values:
+
+| Key | Value | Value Type | Example |
+|-|-|-|-|
+| header | The name of the header you wish to filter on. If this is defined then the named header must explicitly exist on the request otherwise it will be blocked | String | `accept` |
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -344,6 +344,33 @@ docker run --restart=always \
...
+#### Types of Filtering
+
+##### By Header
+
+Add a validation block with the following key/values:
+
+| Key | Value | Value Type | Example |
+|-|-|-|-|
+| header | The name of the header you wish to filter on | String | `accept` |
+| values | If the header exists on the request, its value must match one of the defined strings | Array\<String\> | `["application/vnd.github.v4.sha"]` |
```
carlos-snyk, line 356: Needs updating - header must exist and match one of the defined strings

lib/filters/index.js
```diff
@@ -139,6 +156,12 @@ module.exports = ruleSource => {
         }
       }
 
+      if (headerFilters.length) {
```
carlos-snyk, line 159: very nitpick: the code is expecting an array so `Array.isArray(headerFilters)` is more correct

PR #243, opened 2020-04-20
feat: increase reconnect waiting time

dkontorovskyy:
- [X] Ready for review

#### What does this PR do?
Due to the latest outage, there was an issue with Broker clients reconnecting back to the Broker server. During investigation was uncovered that Broker client will try to reconnect for around 16-18 minutes (re-connect time is randomized).
To improve the user experience during outages, the time of waiting is increased by changing the number of retries and decreasing exponential growth factor.

Previously, we were using the default Primius strategy: 10 retries with exponential growth factor 2.
In this PR, retries are increased to 20 and factor is decreased to 1.5. Which gives us a window between one and one and a half hours to reconnect.
Added logging for each attempt and proximate time till retry.

Added graceful process shutdown to add visibility on the customer side to restart Broker client as their Docker container will be down (after re-running Docker container there will be another 1-1.5h for reconnecting if Broker server is still not up)

### DIFF
--- a/lib/client/socket.js
+++ b/lib/client/socket.js
@@ -27,7 +27,27 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
     },
     pathname : `/primus/${token}`,
   });
-  const io = new Socket(url);
+
+  // Will exponentially back-off from 0.5 seconds to a maximum of 20 minutes
+  // Retry for a total period of around 4.5 hours
+  const io = new Socket(url, {
+    reconnect: {
+      factor: 1.5,
+      retries: 30,
+      max: 20 * 60 * 1000,
+    },
+  });
+
+  io.on('reconnect scheduled', (opts) => {
+    const attemptIn = Math.floor(opts.scheduled / 1000);
+    logger.warn(`Reconnect retry #${opts.attempt} of ${opts.retries} in about ${attemptIn}s`);
+  });
+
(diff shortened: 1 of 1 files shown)


iamtmrobinson:
Given the amount of effort it took CX to contact all customers to restart their brokers manually and the fact that a 90 minute outage doesn't seem beyond the realms of possibility would it be valuable to raise the number of retries and use it in conjunction with a max parameter to limit the backoff? i.e. exponential backoff up to say 20 mins and then retry for longer than 90 mins total.

dkontorovskyy:
@iamtmrobinson I can max check to be 15 or 20 mins and then retry for 3 more hours or so. This will be better?

lib/client/socket.js
```diff
@@ -27,7 +27,25 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
     },
     pathname : `/primus/${token}`,
   });
-  const io = new Socket(url);
+
+  const io = new Socket(url, {
```
iamtmrobinson, line 31: Would potentially add a brief comment here saying something like:
`// Will exponentially back-off from x secs to a maximum of 20 mins, and retry for a total period of around y mins`

dkontorovskyy: Actually added this comment and then removed it :) Pulling it back :)

PR #248, opened 2020-05-21
feat: bump deps (no more lodash)

FauxFaux:
#### What does this PR do?

Upgrading these two deps, which require minimal changes, removes our transitive dependency on `lodash`, and the need for patches.

`snyk-config` upgrade changes it to an exported function, which we can call.
`node-cache` upgrade:
1) removes deprecated api, which we aren't using
2) drops support for node <8, which we don't support

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "lodash.escaperegexp": "^4.1.2",
     "lodash.mapvalues": "^4.6.0",
     "minimatch": "^3.0.4",
-    "node-cache": "^4.2.1",
+    "node-cache": "^5.1.0",
     "path-to-regexp": "^1.8.0",
     "primus": "^6.0.1",
     "primus-emitter": "^3.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,13 +47,13 @@
...
     "minimatch": "^3.0.4",
-    "node-cache": "^4.2.1",
+    "node-cache": "^5.1.0",
     "path-to-regexp": "^1.8.0",
     "primus": "^6.0.1",
     "primus-emitter": "^3.1.1",
     "qs": "^6.9.1",
     "request": "^2.88.0",
-    "snyk-config": "^2.2.3",
+    "snyk-config": "^3.1.1",
```
yuliabaron, line 56: no need in package-lock.json update?

FauxFaux: There's no `package-lock.json` because this is a library.

It doesn't have `package-lock=false` in `.npmrc`, which it probably should have.

PR #253, opened 2020-06-08
chore: remove auth headers to prevent leakage

aviadhahami:
- [x] Ready for review

#### What does this PR do?
When dispatching requests from our broker server to the broker clients, we may relay auth headers. This solves it

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -6,6 +6,7 @@ const uuid = require('uuid/v4');
 const Filters = require('./filters');
 const replace = require('./replace-vars');
 const tryJSONParse = require('./try-json-parse');
+const isObject = require('./is-object');
 const logger = require('./log');
 const version = require('./version');
 const stream = require('stream');
@@ -16,6 +17,10 @@ module.exports = {
   response: responseHandler,
   streamingResponse: streamResponseHandler,
 };
+
+function stringifyIfObject(value) {
+  return isObject(value) ? JSON.stringify(value) : value;
+}
 
 const streams = new NodeCache({
   stdTTL: 18000,  // 5 hours
@@ -197,6 +202,7 @@ function responseHandler(filterRules, config, io) {
         logContext.userAgentHeaderSet = true;
       }
 
+      headers.authorization = {};
       if (result.auth) {
         headers.authorization = result.auth;
         logContext.authHeaderSetByRuleAuth = true;
@@ -270,6 +276,7 @@ function responseHandler(filterRules, config, io) {
       }
 
       logger.debug(logContext, 'sending websocket request over HTTP connection');
+      headers.authorization = stringifyIfObject(headers.authorization);
 
       var req = {
         url: result.url,
--- a/lib/is-object.js
+++ b/lib/is-object.js
@@ -1,0 +1,3 @@
+
+// null is also considered as an object, so we assert truth
+module.exports = (v)=> typeof v === 'object' && !!v;


lib/relay.js
```diff
@@ -6,6 +6,7 @@ const uuid = require('uuid/v4');
 const Filters = require('./filters');
 const replace = require('./replace-vars');
 const tryJSONParse = require('./try-json-parse');
+const isObject = require('lodash.isobject');
```
aviadhahami, line 9: wasn't sure what's better here, if write the `typeof` myself or bring lodash....wdyt?

lib/relay.js
```diff
@@ -270,6 +276,7 @@ function responseHandler(filterRules, config, io) {
       }
 
       logger.debug(logContext, 'sending websocket request over HTTP connection');
+      headers.authorization = stringifyIfObject(headers.authorization);
```
AlbertKogan, line 279: Why you need `stringifyIfObject` if you already doing `headers.authorization = {};`?🧐

aviadhahami: from https://github.com/snyk/broker/blob/fix/broker-auth/lib/relay.js#L206 onward, we MAY modify this object to be a string.

AlbertKogan: ok. But what if you just do `headers.authorization = '{}';` in line [205](https://github.com/snyk/broker/pull/253/files#diff-3a518f5293da3cbae4467a305fb2a077R205)?
What all expected `headers.authorization` values?

iamtmrobinson: I'm not sure I follow, where are you saying we might modify this object to be a string @aviadhahami ? I'd expect the thing that's used to create the req object to be an object.

lib/relay.js
iamtmrobinson: If it's simple enough, I'd prefer doing it ourselves rather than bringing in dependencies.

iamtmrobinson:
@aviadhahami can you help me understand a bit better the situation this is trying to resolve please? What's the required set-up for some sensitive credentials to get leaked and to where do they get leaked?

lib/relay.js
aviadhahami: yup, that's what I thought. Will update

lib/relay.js
aviadhahami: see for example line 222
https://github.com/snyk/broker/blob/fix/broker-auth/lib/relay.js#L222

aviadhahami:
@iamtmrobinson the "vector" is:
A. Generate a broker token (ui)
B. Set up a broker (i.e. artifactory)
C. Set the `ARTIFACTORY_URL` to be something malicious (I used ngrok to simulate)
D. Refresh the broker page (in ui) and see the request hits the ngrok w/ the `SNYK_INTERNAL_BROKER_TOKEN`

Will attach a gif

lib/relay.js
iamtmrobinson: I have a bunch of questions here around things like..
What else can be in this auth object? 
Could we not just set `headers.authorization` to an empty string rather than an object?
Do we know what the contents of result.auth looks like? Is it just a string? Could we just make it one?
Is any of this covered by tests?

aviadhahami: values - IMO it may contain string, null or an object
Tests cover String case for the auth header

lib/relay.js
```diff
@@ -197,6 +202,7 @@ function responseHandler(filterRules, config, io) {
         logContext.userAgentHeaderSet = true;
       }
 
+      headers.authorization = {};
```
darscan, line 205: Why the decision to initialise this with an empty object? `headers` should essentially be `{[key: string]: string}`, surely? Shouldn't this just be `delete headers.authorization`?

aviadhahami: I think it is indeed better than empty object.

Will fix and add tests
(The auth can be a string too, not only am object. I.e `Basic abcd==`)

PR #264, opened 2020-07-22
chore: Convert some functional tests to Jest

robcresswell:
This patch converts a couple of functional tests to Jest, as a follow on to b3322ba7c0657cd74c501f48ee38fd3c84999814

Co-Authored-By: Ohad Finkelstein <ohad.finkelstein@snyk.io>


test/functional/cli-throws.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  afterEach(() => {
+    jest.resetModules();
+  });
+
+  it('when missing broker id', (done) => {
+    expect(() => {
+      cli({ _: ['client'] });
+    }).toThrowError('BROKER_TOKEN');
+
+    done();
```
carlos-snyk, line 13: is the `done()` needed here?

Don't think its doing anything, and when testing promises can do `await expect(...`: https://jestjs.io/docs/en/tutorial-async#asyncawait

robcresswell: @carlos-snyk Ah correct, I don't think its needed; was cherry picking files from another branch and missed this.

I don't think its related to promises though; that code is synchronous (with side effects etc)

robcresswell:
Partially split into #269

PR #272, opened 2020-07-28
chore: convert more test files to Jest

robcresswell:
This is unfortunately not much of a cleanup; we're keeping the tests poorly structured and converting the syntax to Jest. Due to Jest being sensitive to running code, it no longer exits properly so we need to force it with `--forceExit`. TAP does not catch running code in this way.

Also using `as any` a little to allow us to move forward; we could fix the types, but that means moving beyond the test code, and I want to keep the changes small


robcresswell:
Wait, this is nonsense, ignore

PR #275, opened 2020-09-09
feat: added troubleshooting section

AlbertKogan:
- [ ] Ready for review

#### What does this PR do?
A new section that contains some useful workarounds to tackle issues.
This PR adding additional rules to handle Fix PR failures for GitHub / GitHub Enterprise users,
when the root cause is fetch big (> 1MB) manifest file failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,23 @@ Sometime it is required to load sensitive configurations (GitHub/Snyk's token) f
 * Change the workdir of the docker image to be `/broker`/
 Example of such file is located in your broker container at $HOME/.env
 
+### Troubleshooting
+
+#### Support of big manifest files (> 1Mb) for GitHub / GitHub Enterprise
+
+One of the reason for failing of open Fix/Upgrade PRs or PR/recurring tests might be fetching big manifest files (> 1Mb) failure. To address this issue, additional Blob API endpoint should be whitelisted in `accept.json`:
+
+- should be in `private` array
+```json
+{
+    "//": "used to get given manifest file",
+    "method": "GET",
+    "path": "/repos/:owner/:repo/git/blobs/:sha",
+    "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+}
+```
+**Note** To ensure the maximum possible security, we do not enable this rule by default, as usage of this endpoint means that the Snyk platform can theoretically access all files in this repository, as the path does not include specific allowed file names.
+
 ## Misc
 
 * [License: Apache License, Version 2.0](https://github.com/snyk/broker/blob/master/LICENSE)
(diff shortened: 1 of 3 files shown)


iamtmrobinson:
Why does this only apply to GHE?

AlbertKogan:
@iamtmrobinson You are right, need to add this rules for GitHub as well. Thanks!

iamtmrobinson:
Think there should also be test fixture additions/updates to ensure these endpoints work, see [this PR for example](https://github.com/snyk/broker/pull/273/files).

robcresswell (dismissed):

README.md
```diff
@@ -425,6 +425,33 @@ Sometime it is required to load sensitive configurations (GitHub/Snyk's token) f
 * Change the workdir of the docker image to be `/broker`/
 Example of such file is located in your broker container at $HOME/.env
 
+
+
+### Troubleshooting
+
+#### Fail to create Fix PR in GitHub / GitHub Enterprise
+
+One of the reason might be failing of fetching big manifest files (>1 Mb). To address this issue, few additional endpoints should be whitelisted in `accept.json`:
```
carlos-snyk, line 434: This is not only about fix PRs, also about tests. Not being able to fetch a big lock file has many effects.

I'd name this section something like `Need to support files bigger than 1MB in Github / Github enterprise?`

Also, we need to explain why is this not in the default accept.json: because allowing the blob API lets Snyk access any file within the repo.


By the way, the file sha API is not a security problem so we should add it to the default accept.json - less rules to add manually

README.md
```diff
@@ -425,6 +425,22 @@ Sometime it is required to load sensitive configurations (GitHub/Snyk's token) f
...
+
+- should be under `query graphql` section in `valid` array
+```json
+{
+    "//": "query to get file SHA",
+    "path": "query",
+    "regex": "{\\s*repository\\(\\s*owner:\\s*\"[a-zA-Z0-9-_.]+\",\\s*name:\\s*\"[a-zA-Z0-9-_.]+\"\\)\\s*{\\s*object\\(\\s*expression:\\s*\"[a-zA-Z0-9-_./]+:[a-zA-Z0-9-_./]+\"\\)\\s*{\\s*\\.\\.\\.\\s*on\\s*Blob\\s*{\\s*oid\\,\\s*}\\s*}\\s*}\\s*}\\s*"
+}
+```
+**Note**: this endpoint returns file's SHA, that gives Snyk access to all files in given repository.
```
carlos-snyk, line 442: I meant the file sha endpoint is actually safe, we already have the find globs endpoint that returns shas and that is not a security issue.

The security problem is the blobs API (the one that given a sha returns the file contents).

So we should add the file sha to the default rules and put the blobs rule in this document as an optional one.

client-templates/github-com/accept.json.sample
```diff
@@ -1775,6 +1775,12 @@
       "path": "/repos/:owner/:repo/commits",
       "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
     },
+    {
+      "//": "used to get given manifest file",
+      "method": "GET",
+      "path": "/repos/:owner/:repo/git/blobs/:sha",
+      "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+    },
```
carlos-snyk, line 1783: please remove this one and add the graphql endpoint instead (blobs isnt safe)

README.md
AlbertKogan: Oh yes, correct

README.md
```diff
@@ -425,6 +425,23 @@ Sometime it is required to load sensitive configurations (GitHub/Snyk's token) f
...
+- should be in `private` array
+```json
+{
+    "//": "used to get given manifest file",
+    "method": "GET",
+    "path": "/repos/:owner/:repo/git/blobs/:sha",
+    "origin": "https://${GITHUB_TOKEN}@${GITHUB_API}"
+}
+```
+**Note**: this endpoint using SHA of file to return file's content. Together with whitelisted GraphQL Blob query (`query to get file SHA`) it gives Snyk access to all files in given repository.
```
robcresswell, line 443: Small suggestion on the wording:

**Note** To ensure the maximum possible security, we do not enable this rule by default, as usage of this endpoint means that the Snyk platform can theoretically access all files in this repository, as the path does not include specific allowed file names.

robcresswell (approved):

carlos-snyk (approved):

PR #278, opened 2020-09-24
feat: Azure Repos support

AlbertKogan:
- [ ] Ready for review

#### What does this PR do?
Adding support of Azure Repos integration.
New files:
- client-templates/azure-repos/.env.sample - template of required settings for run broker client
- client-templates/azure-repos/accept.json.sample - whitelist template of required Azure API endpoints
- dockerfiles/azure-repos/Dockerfile - Dockerfile template

Additional changes:
- adds Azure Repos to sanitisation logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,46 @@ ENV BROKER_CLIENT_URL   http://my.broker.client:8000
 ENV PORT                8000
 ```
 
-### Artifactory 
+### Azure Repos
+
+To use the Broker client with [Azure](https://azure.microsoft.com/en-us/services/devops/), run `docker pull snyk/broker:azure-repos` tag. The following environment variables are mandatory to configure the Broker client:
+
+- `BROKER_TOKEN` - the Snyk Broker token, obtained from your Azure Repos integration settings view (app.snyk.io).
+- `AZURE_REPOS_TOKEN` - an Azure Repos personal access token. [Guide](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page) how to get/create the token. Required scopes: ensure Custom defined is selected and under Code select _Read & write_
+- `AZURE_REPOS_ORG` - organization name, which can be found in your Organization Overview page in Azure
+- `PORT` - the local port at which the Broker client accepts connections. Default is 7341.
+- `BROKER_CLIENT_URL` - the full URL of the Broker client as it will be accessible by your Azure Repos' webhooks, such as `http://my.broker.client:7341`
+
+#### Command-line arguments
+
+You can run the docker container by providing the relevant configuration:
+
+```
+docker run --restart=always \
+           -p 8000:8000 \
+           -e BROKER_TOKEN=secret-broker-token \
+           -e AZURE_REPOS_TOKEN=secret-azure-token \
+           -e AZURE_REPOS_ORG=org-name \
+           -e BROKER_CLIENT_URL=http://my.broker.client:8000 \
+           -e PORT=8000 \
+       snyk/broker:azure-repos
+```
+
+#### Derived docker image
+
+Another option is to build your own docker image and override relevant environment variables:
+
+```
+FROM snyk/broker:azure-repos
+
+ENV BROKER_TOKEN        secret-broker-token
+ENV AZURE_REPOS_TOKEN   secret-azure-token
+ENV AZURE_REPOS_ORG      org-name
+ENV BROKER_CLIENT_URL   http://my.broker.client:8000
+ENV PORT                8000
+```
+
+### Artifactory
 
 To use the Broker client with an artifactory deployment, run `docker pull snyk/broker:artifactory` tag. The following environment variables are needed to customize the Broker client:
 
--- a/client-templates/azure-repos/.env.sample
+++ b/client-templates/azure-repos/.env.sample
@@ -1,0 +1,28 @@
+# Guide how to get/create the token https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page
+# Scopes: Ensure Custom defined is selected and under Code select Read & write
+AZURE_REPOS_TOKEN=<token>
+
+# https://dev.azure.com/<organisation name>
+AZURE_REPOS_ORG=<organisation name>
+
+# the host excluding scheme
+AZURE_REPOS_HOST=dev.azure.com
+
+# your unique broker identifier
+BROKER_TOKEN=<broker-token>
+
+# the url of your broker client (including scheme and port)
+BROKER_CLIENT_URL=https://<broker.client.hostname>:<port>
+
+# auth header with special format
+BROKER_CLIENT_VALIDATION_BASIC_AUTH="PAT:$AZURE_REPOS_TOKEN"
+
+# The URL of the Snyk broker server
+BROKER_SERVER_URL=https://broker.snyk.io
+
+# the fine detail accept rules that allow Snyk to make API requests to your
+# bitbucket server instance
+ACCEPT=accept.json
+
+# The path for the broker's internal healthcheck URL. Must start with a '/'.
+BROKER_HEALTHCHECK_PATH=/healthcheck
--- a/client-templates/azure-repos/accept.json.sample
+++ b/client-templates/azure-repos/accept.json.sample
@@ -1,0 +1,242 @@
+{
+  "public": [
+    {
+      "//": "used for pushing up webhooks from Azure",
+      "method": "POST",
+      "path": "/webhook/azure-repos/:webhookId"
+    }
+  ],
+  "private": [
+    {
+      "//": "get list of projects for given organisation",
+      "method": "GET",
+      "path": "/_apis/projects",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "get specific repository for given organisation",
+      "method": "GET",
+      "path": "/:owner/_apis/git/repositories/:repo",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "get list of repositories for given organisation",
+      "method": "GET",
+      "path": "/:owner/_apis/git/repositories",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "get list of refs",
+      "method": "GET",
+      "path": "/:owner/_apis/git/repositories/:repo/refs",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "search through repositories of given organisation",
+      "method": "GET",
+      "path": "_apis/git/repositories",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "create hook",
+      "method": "POST",
+      "path": "/_apis/hooks/subscriptions",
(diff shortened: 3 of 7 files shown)


lib/auth-header.js
```diff
@@ -7,10 +7,15 @@ module.exports = ({ scheme, token, username, password }) => {
...
-      replace(username, config),
-      replace(password, config),
-    ].join(':');
+    let basicAuth;
+
+    if (token) {
+      basicAuth = replace(token, config);
+    } else {
+      basicAuth = [replace(username, config), replace(password, config)].join(
+        ':',
```
AlbertKogan, line 16: Automatic formatting looks a bit weird here

lib/auth-header.js
```diff
@@ -7,10 +7,15 @@ module.exports = ({ scheme, token, username, password }) => {
...
-    const basicAuth = [
-      replace(username, config),
-      replace(password, config),
-    ].join(':');
+    let basicAuth;
+
+    if (token) {
+      basicAuth = replace(token, config);
+    } else {
+      basicAuth = [replace(username, config), replace(password, config)].join(
```
carlos-snyk, line 15: The token part doesn't look right. Basic auth is supposed to be username and password.

https://en.wikipedia.org/wiki/Basic_access_authentication

> In the context of an HTTP transaction, basic access authentication is a method for an HTTP user agent (e.g. a web browser) to provide a user name and password when making a request. In basic HTTP authentication, a request contains a header field in the form of Authorization: Basic <credentials>, where credentials is the Base64 encoding of ID and password joined by a single colon :.

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,222 @@
+{
+  "public": [
+    {
+      "//": "used for pushing up webhooks from Azure",
+      "method": "POST",
+      "path": "/webhook/azure-repos/webhookId"
```
carlos-snyk, line 6: Suspect `webhookId` is a variable here:
```suggestion
      "path": "/webhook/azure-repos/:webhookId"
```

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,222 @@
...
+    }
+  ],
+  "private": [
+    {
+      "//": "get list of projects for given organisation",
+      "method": "GET",
+      "path": "/_apis/projects",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZUR_REPOS_ORG}",
```
carlos-snyk, line 14:
```suggestion
      "origin": "https://${AZURE_REPOS_HOST}/${AZURE_REPOS_ORG}",
```

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,222 @@
...
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
```
carlos-snyk, line 17: Could do 
```suggestion
        "username": "PAT",
        "password": "${AZURE_REPOS_TOKEN}"
```

This way the basic auth header would work without any change

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,222 @@
...
+          ]
+        }
+      ],
+      "auth": {
+        "scheme": "basic",
+        "token": "${BROKER_CLIENT_VALIDATION_BASIC_AUTH}"
+      }
+    },
+    {
+      "//": "get list of globs for given repository",
```
carlos-snyk, line 146: Should use customer-friendly terms, in this case `files` instead of `globs`

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,222 @@
...
+      "method": "GET",
+      "path": "/:owner/_apis/git/repositories/:repo/items",
+      "origin": "https://${AZURE_REPOS_HOST}/${AZUR_REPOS_ORG}",
+      "valid": [
+        {
+          "queryParam": "recursionLevel",
+          "values": ["full"]
+        }
+      ],
```
carlos-snyk, line 155: if we remove this section, is every query param valid? Don't see the point in restricting query params when no param is a security issue.

I mean, if Snyk requested a single level instead of full it's not a security issue, so the Broker rules shouldn't be concerned by that. 

It would also let Snyk change the recursion in the future without changes to Broker.

client-templates/azure-repos/accept.json.sample
AlbertKogan: Yes, you are right. But it seems for me a bit confusing. `PAT` is defiantly not a username. I prefer to find better way

lib/auth-header.js
AlbertKogan: Agree. But Azure reimplement authentication in its own way

client-templates/azure-repos/accept.json.sample
AlbertKogan: This is interesting one. 

Since Azure using same endpoint for fetching globs (paths) and getting file content, we need somehow restrict access to user's files. 
If we remove `valid ` array here, it basically will mean that we can pass any set of query parameters. For example: 
`?recursionLevel=full&download=true` that means fetch all files in repo. To avoid this behaviour we need to pass granular set of query parameters, that will exclude combinations of `recursionLevel=full` and `download=true`

client-templates/azure-repos/accept.json.sample
```diff
@@ -0,0 +1,242 @@
...
+            "**/project.assets.json",
+            "**%2Fproject.assets.json",
+            "**/Podfile",
+            "**%2FPodfile",
+            "**/Podfile.lock",
+            "**%2FPodfile.lock"
+          ]
+        },
+        {
+          "queryParam": "recursionLevel",
```
AlbertKogan, line 140: @carlos-snyk Add more query parameters to restrict access to repo's files

dockerfiles/azure-repos/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# NOTA BENE: without `/` character in the end of URL
+ENV AZURE_REPOS_HOST dev.azure.com
+
+ENV BROKER_CLIENT_VALIDATION_BASIC_AUTH "PAT:$AZURE_REPOS_TOKEN"
+
+ENV AZURE_REPOS_ORG <azure-repos-org>
+
+ENV BROKER_SERVER_URL https://broker.snyk.io
+
+ENV ACCEPT accept.json
```
yuliabaron, line 35: I don't see this ENV var in other Dockerfiles, same is true for some other ENVs

lib/auth-header.js
yuliabaron: You might want to add a comment which SCMs use token

lib/auth-header.js
robcresswell: A template literal would probably look neater here?

```typescript
basicAuth = `${replace(username, config)}:${replace(password, config)}` 
```

dockerfiles/azure-repos/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+ENV BROKER_SERVER_URL https://broker.snyk.io
+ENV BROKER_HEALTHCHECK_PATH /healthcheck
+
+# The port used by the broker client to accept internal connections
+# Default value is 7341
+# ENV PORT 7341
+
+# The URL of your broker client (including scheme and port)
+# This will be used as the webhook payload URL 
+# ENV BROKER_CLIENT_URL http://<broker.client.hostname>:$PORT
```
yuliabaron, line 42: why do you comment this one out and the upper ones not? the user will need to define this one in the same manner as others

dockerfiles/azure-repos/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# Guide how to get/create the token https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page
+# Scopes: Ensure Custom defined is selected and under Code select Read & write
+ENV AZURE_REPOS_TOKEN <azure-repos-token>
+# NOTA BENE: without `/` character in the end of URL
+ENV AZURE_REPOS_HOST dev.azure.com
+
+# Your unique Broker identifier
+ENV BROKER_TOKEN <broker-token>
+ENV BROKER_CLIENT_VALIDATION_BASIC_AUTH "PAT:$AZURE_REPOS_TOKEN"
+ENV BROKER_SERVER_URL https://broker.snyk.io
```
yuliabaron, line 33: this one and the next one should not be here - those are defaults in .env file, and are not custom-based. You can move them up, before line 16, or remove them

AlbertKogan: You are right. No need to duplicate this values here

yuliabaron (dismissed):

client-templates/azure-repos/accept.json.sample
iamtmrobinson: I agree with @AlbertKogan here, let's keep it explicit rather than hacking around the existing logic.

lib/auth-header.js
```diff
@@ -7,10 +7,10 @@ module.exports = ({ scheme, token, username, password }) => {
...
   if (scheme === 'basic') {
-    const basicAuth = [
-      replace(username, config),
-      replace(password, config),
-    ].join(':');
-    return `Basic ${Buffer.from(basicAuth).toString('base64')}`;
+    return `Basic ${Buffer.from(
+      token
+        ? replace(token, config) // Not standardized Basic auth format, currently using in Azure Repos API
+        : `${replace(username, config)}:${replace(password, config)}`,
```
iamtmrobinson, line 13: I think this would be more readable if we extracted things so the comment and ternary don't muddy the logic as much e.g.
```
const basicAuth = token ? getTokenBasedBasicAuth(token, config) : `${replace(username, config)}:${replace(password, config)}`;
return `Basic ${basicAuth}`;
```

Option to also move the basic auth to a separate function, though it didn't seem to add much.

AlbertKogan: Not sure why we need `getTokenBasedBasicAuth` since it is one line: `replace(token, config)`

iamtmrobinson: It doesn't make a huge difference, but it's quicker to process a readable function name than some logic and a comment i.e. a ternary with `getTokenBasedBasicAuth() : getBasicAuth()` is quicker to grok than interpreting the logic that they constitute, let alone the comment which contains info that isn't always immediately necessary.

README.md
```diff
@@ -181,6 +181,45 @@ ENV BROKER_CLIENT_URL   http://my.broker.client:8000
 ENV PORT                8000
 ```
 
+### Azure Repos
+
+To use the Broker client with [Azure](https://azure.microsoft.com/en-us/services/devops/), run `docker pull snyk/broker:azure-repos` tag. The following environment variables are mandatory to configure the Broker client:
+
+- `BROKER_TOKEN` - the Snyk Broker token, obtained from your Azure Repos integration settings view (app.snyk.io).
+- `AZURE_REPOS_TOKEN` - an Azure Repos personal access token
```
iamtmrobinson, line 189: Do we need to define what permissions are required here? We seem to for other SCMs

README.md
```diff
@@ -181,6 +181,45 @@ ENV BROKER_CLIENT_URL   http://my.broker.client:8000
...
+- `AZURE_REPOS_ORG` - organization name, could be found in Organization Overview page
```
iamtmrobinson, line 190:
```suggestion
- `AZURE_REPOS_ORG` - organization name, which can be found in your Organization Overview page in Azure
```

README.md
```diff
@@ -181,6 +181,45 @@ ENV BROKER_CLIENT_URL   http://my.broker.client:8000
...
+- `PORT` - the local port at which the Broker client accepts connections. Default is 7341.
+- `BROKER_CLIENT_URL` - the full URL of the Broker client as it will be accessible by Azure Repos webhooks, such as `http://my.broker.client:7341`
```
iamtmrobinson, line 192:
```suggestion
- `BROKER_CLIENT_URL` - the full URL of the Broker client as it will be accessible by your Azure Repos' webhooks, such as `http://my.broker.client:7341`
```

README.md
AlbertKogan: Mentioned this in Dockerfile. Worth to add here as well

client-templates/azure-repos/.env.sample
```diff
@@ -0,0 +1,28 @@
...
+AZURE_REPOS_ORG=<organisation name>
+
+# the host excluding scheme
+AZURE_REPOS_HOST=dev.azure.com
+
+# your unique broker identifier
+BROKER_TOKEN=<broker-token>
+
+# the url of your broker client (including scheme and port)
+BROKER_CLIENT_URL=<http://<broker client hostname:port>
```
iamtmrobinson, line 15:
```suggestion
# BROKER_CLIENT_URL=
```
This suggestion is purely because it then matches what other samples have

AlbertKogan: @yuliabaron suggested opposite, uncomment this😅 From user perspective, things that are commented out seems not important/must. But this is required setting

iamtmrobinson: Sounds fair, but can we raise a separate PR at the same time that updates all the other samples to match please?
I'd suggest the following..
`BROKER_CLIENT_URL=https://<BROKER_CLIENT_HOSTNAME>:<PORT>`

AlbertKogan: Sure. Wil do

iamtmrobinson (dismissed):

iamtmrobinson (approved):

PR #285, opened 2020-11-02
feat: add support for binary chunked data and body string substitution

orsagie:
- [x] Ready for review

#### What does this PR do?
Support substitution in streamed and non streamed response bodies in the client. For Artifactory, this is necessary to support NPM calls.


lib/filters.js
```diff
@@ -3,7 +3,7 @@ const pathRegexp = require('path-to-regexp');
 const qs = require('qs');
 const path = require('path');
 const undefsafe = require('undefsafe');
-const replace = require('./replace-vars');
+const { replace } = require('./replace-vars');
```
AlbertKogan, line 6: Wondering how it works🧐

client-templates/artifactory/.env.sample
```diff
@@ -13,3 +14,6 @@ ACCEPT=accept.json
 
 # The path for the broker's internal healthcheck URL. Must start with a '/'.
 BROKER_HEALTHCHECK_PATH=/healthcheck
+
+# Provide RES_BODY_URL_SUB with the URL of the artifactory without credentials and http protocol
```
AlbertKogan, line 18: Not sure that description is clear. Can you elaborate a bit more about `RES_BODY_URL_SUB `? Why we need it?

lib/replace-vars.js
```diff
@@ -1,6 +1,33 @@
...
+function replace(input, source) {
   return (input || '').replace(/(\${.*?})/g, (_, match) => {
     const key = match.slice(2, -1); // ditch the wrappers
     return source[key] || '';
   });
-};
+}
+
+function replaceUrlPartialChunk(chunk, prevPartial, config) {
+  const replaceRegExp = new RegExp(config.RES_BODY_URL_SUB, 'g');
```
AlbertKogan, line 14: Think it's OK to import `config` here instead passing as a parameter.

client-templates/artifactory/.env.sample
orsagie: Why we need it - when npm does an install it will first grab metadata about the package. This metadata includes where the tarball for each version is located. The url location for the tarball will be the artifactory url, which is inaccessible to us, and needs to be substituted to the broker url base.

lib/replace-vars.js
orsagie: Should relay.js also import config?

orsagie: This is config is created and then passed through the broker, see: https://github.com/snyk/broker/blob/551f750137c39b74922245b7c00613c576525b9f/lib/index.js#L20

client-templates/artifactory/.env.sample
orsagie: Should this variable be renamed or only the description?

client-templates/artifactory/accept.json.sample
```diff
@@ -2,7 +2,8 @@
   "public": [
     {
       "method": "any",
-      "path": "/*"
+      "path": "/*",
+      "stream": true
```
orsagie, line 6: Streaming is required for artifactory, as large chunks of binary data will be passed through

client-templates/artifactory/.env.sample
AlbertKogan: @orsagie Thanks for clarifying this☺️ I think it might be a good idea to update description. Don't have any suggestions on naming🤷‍♂️

AlbertKogan (approved):

admons (approved):

client-templates/artifactory/.env.sample
orsagie: updated the description

AlbertKogan (approved):

PR #286, opened 2020-11-04
test: github big file rule

carlos-snyk:
- [x] Ready for review

#### What does this PR do?

Adds a test for a rule that allows getting file shas via the github graphql API - the file sha is used as a parameter for the blobs (big files) API.

The rule is not part of the default accept.json but useful for some Broker users that need to support big files.


AlbertKogan (approved):
💫

PR #289, opened 2020-11-08
fix: make encoded ghe accept json /dockerfile path valid

ohad2712:
A [customer ticket](https://snyk.zendesk.com/agent/tickets/7256) has lead to this potential solution where the current GHE `accept.json` template has suggested an invalid uri for scanning a Dockerfile with an encoded path.
This PR removes the extra irrelevant `2F` from the rule's uri path.

### DIFF
--- a/test/fixtures/accept/ghe.json
+++ b/test/fixtures/accept/ghe.json
@@ -690,6 +690,18 @@
     {
       "//": "used to update manifest or lock",
       "method": "PUT",
+      "path": "/:name/:repo/:path*/package.json",
+      "origin": "https://${GITHUB_TOKEN}@${GITHUB_RAW}"
+    },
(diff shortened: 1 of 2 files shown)


client-templates/github-enterprise/accept.json.sample
```diff
@@ -1008,7 +1008,7 @@
     {
       "//": "used to scan Dockerfile",
       "method": "GET",
-      "path": "/repos/:name/:repo/contents/:path*%2F2FDockerfile",
+      "path": "/repos/:name/:repo/contents/:path*%2FDockerfile",
```
AlbertKogan, line 1011: Nice catch @ohad2712 !

AlbertKogan (approved):

PR #292, opened 2020-11-10
feat: Support PR statuses for brokered Azure Repos

robcresswell:
This patch updates the Azure Repos sample accept.json to include a new rule that allows post PR statuses. Until now, Snyk only supported commit statuses, which are different from PR statuses for Azure Repos

See https://github.com/snyk/azure-repos-agent/pull/219 for the implementation of this endpoint in the Azure Repos agent

### DIFF
--- a/client-templates/azure-repos/accept.json.sample
+++ b/client-templates/azure-repos/accept.json.sample
@@ -199,6 +199,16 @@
       }
     },
     {
+      "//": "update status of given pull request",
+      "method": "POST",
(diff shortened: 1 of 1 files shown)


iamtmrobinson:
Are there any rules that should be removed as part of this if this replaces them?

robcresswell:
@iamtmrobinson Nope, existing rules are staying. This is extra, not replacing.

iamtmrobinson (approved):

PR #295, opened 2020-12-11
feat: Add container registry agent support

jurglic:
Add broker support for container-registry-agent by providing:
* template: accept.json, env
* Dockerfile to build image
* README section

- [x] Ready for review

### DIFF
--- a/client-templates/container-registry-agent/accept.json.sample
+++ b/client-templates/container-registry-agent/accept.json.sample
@@ -1,0 +1,62 @@
+{
+  "public": [
+  {
+    "//": "image scan callback (v1)",
+    "method": "POST",
+      "path": "/api/v1/import/app"
+  },
+  {
+    "//": "image scan callback (v2)",
+    "method": "POST",
+    "path": "/api/v2/import/done"
+  },
+  {
+    "//": "image scan callback (v2)",
+    "method": "POST",
+    "path": "/api/v2/import/result"
+  }
+  ],
+  "private": [
+  {
+    "method": "GET",
+    "path": "/healthcheck",
+    "origin": "${CR_AGENT_URL}"
+  },
+  {
+    "//": "validates credentials with connection to container registry",
+    "method": "GET",
+    "path": "/validate-credentials",
+    "origin": "${CR_AGENT_URL}"
+  },
+  {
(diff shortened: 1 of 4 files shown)


client-templates/container-registry-agent/accept.json.sample
```diff
@@ -0,0 +1,48 @@
...
+    "origin": "${CR_AGENT_URL}"
+  },
+  {
+    "method": "GET",
+    "path": "/get-state-identifier",
+    "origin": "${CR_AGENT_URL}"
+  },
+  {
+    "method": "GET",
+    "path": "/list*",
```
snaftaly, line 34: Looking at other broker client examples, I think you can split this into two entries, one with the path: "/list" (for listing repos) and another one with the path "/list/:repo" (for listing tags).

client-templates/container-registry-agent/accept.json.sample
```diff
@@ -0,0 +1,48 @@
...
+  },
+  {
+    "method": "POST",
+    "path": "/api/v2/import/result"
+  }
+  ],
+  "private": [
+  {
+    "method": "GET",
+    "path": "/healthcheck",
```
snaftaly, line 19: I'm not sure this is needed, as the registry will not send healthchecks to the remote DRA. wdyt?

client-templates/container-registry-agent/accept.json.sample
```diff
@@ -0,0 +1,48 @@
+{
```
snaftaly, line 1: A general comment - consider adding a "//" comment field in each endpoint, to explain its purpose.

client-templates/container-registry-agent/accept.json.sample
jurglic: Not strictly necessary atm. But imho good to have this open from the get go, if we want to monitor it in the future.

client-templates/container-registry-agent/accept.json.sample
jurglic: Updated

client-templates/container-registry-agent/accept.json.sample
jurglic: Done

iamtmrobinson (approved):

PR #296, opened 2020-12-13
feat: support proxy authentication

ohad2712:
This PR handles the proper env vars for a proxy used in broker client, and basically adds the support for any proxy authentication in broker.

### DIFF
--- a/lib/patch-https-request-for-proxying.js
+++ b/lib/patch-https-request-for-proxying.js
@@ -7,7 +7,7 @@
  */
 const url = require('url');
 const tunnel = require('tunnel');
-const { brokerServerUrl, httpsProxy, noProxy } = require('./config');
+const { brokerServerUrl, httpsProxy, noProxy, proxyAuth } = require('./config');
 const brokerServer = url.parse(brokerServerUrl || '');
 brokerServer.port =
   brokerServer.port || brokerServer.protocol === 'https:' ? '443' : '80';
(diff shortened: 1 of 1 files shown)


lib/patch-https-request-for-proxying.js
```diff
@@ -80,11 +80,10 @@ function shouldProxy(uri) {
 // Entry point: To patch or not to patch?
 if (brokerServer.host && httpsProxy && shouldProxy(brokerServer)) {
   const proxy = url.parse(httpsProxy);
+  let tunnelProxy = ({ host: hostname, port } = proxy);
+  tunnelProxy = { ...tunnelProxy, ...(proxyAuth && { proxyAuth }) };
```
dkontorovskyy, line 84:
```suggestion
  const {host: hostname, port} = url.parse(httpsProxy);
  const tunnelProxy = { ...{ host, port }, ...{proxyAuth} };
```
And make `proxyAuth` by default to return `{}`

ohad2712: Shouldn't we just not add this property if it's not defined?

dkontorovskyy: fixed suggestion

ohad2712: It's similar to what I wrote. It should behave exactly as you intended.

dkontorovskyy: I agree, but it's much more readable.

iamtmrobinson: I agree with @dkontorovskyy here, I find the suggested change more readable.

carlos-snyk (approved):

PR #297, opened 2020-12-16
fix: wrap websocket response handler to prevent broker crashing

AlbertKogan:
This patch introduces wrapping sending websocket response back with `try..catch`.
This will help us mitigate the problem with unexpected error:
`TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of
type string or an instance of Buffer or Uint8Array. Received an instance of Object`

- [ ] Ready for review

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -150,14 +150,23 @@ function requestHandler(filterRules) {
             .status(ioResponse.status)
             .set(ioResponse.headers);
 
-          // keep chunked http requests without content-length header
-          const isChunked =
-            undefsafe(ioResponse, 'headers.transfer-encoding') === 'chunked';
-          if (isChunked) {
-            httpResponse.write(ioResponse.body);
-            httpResponse.end();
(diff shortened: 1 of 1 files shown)


lib/relay.js
```diff
@@ -150,14 +150,23 @@ function requestHandler(filterRules) {
...
+            if (encodingType === 'chunked') {
+              httpResponse.write(ioResponse.body);
+              httpResponse.end();
+            } else {
+              httpResponse.send(ioResponse.body);
+            }
+          } catch (err) {
+            logger.error(
+              { ...logContext, encodingType, err },
+              'Unsupported encoding type',
```
robcresswell, line 168: We don't know what causes the error, this should just be something like  `'error forwarding response`

robcresswell (dismissed):
Approving to unblock, but update the err message then merge please 😃

robcresswell (approved):

PR #301, opened 2020-12-30
fix: artifactory systmecheck endpoint

AlbertKogan:
- [ ] Ready for review

#### What does this PR do?

This patch fixing 406 error of `/api/system/ping` endpoint. It returns `text/plain` response
and not allow to pass `application/json` header.

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -49,6 +49,7 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
       config.brokerClientValidationMethod || 'GET';
     const brokerClientValidationTimeoutMs =
       config.brokerClientValidationTimeoutMs || 5000;
+    const isJsonResponse = !config.brokerClientValidationJsonDisabled;
 
     const data = {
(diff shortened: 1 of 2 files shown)


lib/client/index.js
```diff
@@ -49,6 +49,7 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
       config.brokerClientValidationMethod || 'GET';
     const brokerClientValidationTimeoutMs =
       config.brokerClientValidationTimeoutMs || 5000;
+    const isJsonResponse = !config.brokerClientValidationJsonDisabled;
```
ohad2712, line 52: I'd call it `...ENABLED` then, and not `...DISABLED`. Right? 🤔

ohad2712 (approved):

lib/client/index.js
AlbertKogan: The point is to turn it off only for artifactory integration. If we will do `_ENABLED`, we need to update all .env files for all integrations...

PR #307, opened 2021-01-26
feat: update all agent dockerfiles with new remediated image

ohad2712:
Updates all dockerfiles to use base image with a reduced vuln count.

The update includes a new much remediated image (built on top of an ubuntu image) vs. the original image in use. We assessed the vulns in both images and determined they can not be exploited when used to run the broker.

image (12)

### DIFF
--- a/dockerfiles/artifactory/Dockerfile
+++ b/dockerfiles/artifactory/Dockerfile
@@ -1,19 +1,35 @@
-FROM node:12-buster-slim
+FROM snyk/ubuntu as base
 
 MAINTAINER Snyk Ltd
 
-RUN npm install --global snyk-broker
+USER root
+
+RUN apt update && apt install -y make python gcc
 
 RUN apt-get update && apt-get install -y ca-certificates
 
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+
+ENV PATH=$PATH:/home/node/.npm-global/bin 
+
+RUN npm install --global snyk-broker 
+
+
+
+FROM snyk/ubuntu
+
+ENV PATH=$PATH:/home/node/.npm-global/bin 
+
+COPY --from=base /home/node/.npm-global /home/node/.npm-global
+
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
 # Don't run as root
 WORKDIR /home/node
-USER node
+USER node 
 
 # Generate default accept filter
 RUN broker init artifactory
-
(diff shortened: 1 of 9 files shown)


dockerfiles/artifactory/Dockerfile
```diff
@@ -1,19 +1,33 @@
...
-USER node
+FROM snyk/ubuntu
 
-# Generate default accept filter
-RUN broker init artifactory
+ENV PATH=$PATH:/home/node/.npm-global/bin 
+
+COPY --from=base /home/node/.npm-global /home/node/.npm-global
 
+USER root
```
iamtmrobinson, line 27: Previously we explicitly called out `# Don't run as root`, do we want to do the same here?

dockerfiles/artifactory/Dockerfile
```diff
@@ -1,19 +1,33 @@
...
+
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+
+ENV PATH=$PATH:/home/node/.npm-global/bin 
+
 RUN npm install --global snyk-broker
 
+USER root 
+
 RUN apt-get update && apt-get install -y ca-certificates
```
AlbertKogan, line 19: Just want to verify that this is done by purpose. We have `RUN apt update` on line 7. Can we make certificate installation there as well?

iamtmrobinson (approved):

dockerfiles/artifactory/Dockerfile
ohad2712: This is on purpose, as this is the idea of using a [multistage build in docker](https://docs.docker.com/develop/develop-images/multistage-build/).
But this question did help me think that I accidentally put the certs installation before the second phase (`FROM`), when it should be after, so that it'll be in the final container, so nice catch anyway.

dockerfiles/artifactory/Dockerfile
ohad2712: So fixed it. You'll still see the script running as `root` in the first stage, but don't worry about it.
These dockerfiles use what's called a [multistage build in docker](https://docs.docker.com/develop/develop-images/multistage-build/). The first phase, which installs everything has to be when running as `root`, but this is just the build stage.
A few lines later, we start from scratch (from a clean `snyk/ubuntu` image to be exact), and just copy over everything we installed before. In this (final) stage, the user is no longer `root`, but `node`, which is the only thing that matters for the final container that is deployed.

yeforriak (approved):

PR #327, opened 2021-06-09
Feat: support git client

shinin-light:
- [x] Ready for review

#### What does this PR do?
As part of the efforts to allow on-prem customers to have Snyk Code analysis results, we've built a support service that can be run alongside the Broker Client to clone the repositories and upload the code to our servers to be analyzed. In order to support this solution, we've updated the Broker Client to be aware of this new architecture and redirect all the requests to a specific endpoint towards the new snyk git client.

The expected behaviour at the moment is:
- a new `GIT_CLIENT_URL` variable is introduced in the `.env` file to be specified by the customer during the deployment, alongside three support variables `GIT_URL`, `GIT_USERNAME`, `GIT_PASSWORD` that are just used to capture under a common name different SCM-related variables that already present.
- every request towards `/snykgit/*` is redirected to the relative `/*` path on the git client.
- if a an analysis key is present in the body of the aforementioned request, the proper scm address and credentials are injected in the body before the forwarding.
- no change is required for the Snyk Broker Server, but beta customers will need to update their broker client version and configuration together with deploying the new git client to benefit from on-prem code analyses.

Possible future scenarios:
- we expect other products and teams to start using the snyk git client in the future after the communication protocol is established. For this reasons we want the broker changes to be simple and allow us a variety of approaches end-to-end. 

There are some design choices that may be useful to review for maintainability and longevity of the product:
- use of a path specification to specify the direction of the forwarding (scm vs git-client). This could also be achieved through a custom header, if that's advisable.
- use of env variables to have a common scm-agnostic name reference to info that is already present elsewhere in the config (e.g. GITHUB_TOKEN => GIT_PASSWORD)
- the only env variable that has been added to the dockerfiles examples is the `GIT_CLIENT_URL`. Do we have to add also the support variables?
- all the traffic that should be forwarded to the git client is managed by the code in `relay-git.js`. At the moment this file looks just like a simplified version of the `relay.js` code, so maybe some refactoring may be done in a future iteration.  

- We did test the code locally and added some tests for the new functionality, however the end-to-end interaction with real SCMs has not being tested at all (all the e2e tests have been performed for a brokered github.com integration).

We had multiple design iterations for possible on-prem analysis flow, all of which included some changes to the broker client. If you want to know more, check out the other [proposed design here](https://docs.google.com/presentation/d/16U72rCxbrTO_lCBqU6FeNDwpzkPvwsXvJGncjF8PmtM/edit?usp=sharing) or chat with us at [#feat-code-access-2021](https://snyk.slack.com/archives/C01T8J66QMS)

#### Screenshots
image


lib/relay.js
```diff
@@ -186,6 +187,10 @@ function responseHandler(filterRules, config, io) {
     { url, headers = {}, method, body = null, streamingID = '' } = {},
     emit,
   ) => {
+    if (url && url.startsWith(relayGit.GIT_CLIENT_REDIRECT_PATH)) {
+      return relayGit.response(config, io)(brokerToken)({ url, headers, method, body, streamingID }, emit);
```
neonnoon, line 191: Question: just wondering: since `relayGit.response` is used only here and called directly, why did you go for a curried function call? Probably reasonable to simplify future refactoring between `relay.js`/`relay-git.js`?

lib/relay-git.js
```diff
@@ -0,0 +1,115 @@
...
+
+    const { GIT_CLIENT_URL, GIT_URL, GIT_USERNAME, GIT_PASSWORD } = config;
+    let errorStatus;
+    if (brokerToken) errorStatus = 'NO_GIT_CLIENT_ON_BROKER_SERVER';
+    else if (!GIT_CLIENT_URL) errorStatus = 'MISSING_GIT_CLIENT_URL';
+    else if (!GIT_URL || !GIT_USERNAME || !GIT_PASSWORD) errorStatus = 'MISSING_GIT_SERVER_CONFIG';
+    if (errorStatus) {
+      logger.error(
+        { ...logContext, errorStatus },
+        'received git client request with wronng configuration'
```
neonnoon, line 44: Typo: `wronng`

lib/relay-git.js
```diff
@@ -0,0 +1,115 @@
...
+        status: 500,
+        body: errorStatus,
+      });
+    }
+
+    // injecting scm url and credentials in the request body
+    if (body) {
+      const parsedBody = tryJSONParse(body);
+      if (parsedBody && parsedBody.key && parsedBody.key.gitURI) {
+        parsedBody.key.gitURI = parsedBody.key.gitURI.includes('://') ?
```
neonnoon, line 56: Question: This seems a bit shaky to me, can't we make a stronger assumption about `gitURI` here?

neonnoon (approved):
Nice description and explanation with the flow diagram. Looks good to me. Just a couple of minor questions.

lib/relay-git.js
shinin-light: Another good point. In both the git-client and DeepCode's API-server, we use `joi` to check the exact format of the `gitURI`, so the broker output value **must** be `<schema>://<scm>/<something-something>`. Now the input on the other end is controlled by registry. Right now that we only use this for Code Access, I'm ok with sending a fake `<schema>://<fake-scm>` in the request to be converted by the Broker with the correct address, but in future we might want to change it (or other products may find it obnoxious and prefer to just output `/<owner>/<repo>.git`). So the idea here is to be a bit loose on the request and tolerant with the format to leave a bit of space for further improvements without having to deploy new Broker Clients to every customer and deal with payload versioning.

lib/relay-git.js
```diff
@@ -0,0 +1,115 @@
...
+    if (brokerToken) errorStatus = 'NO_GIT_CLIENT_ON_BROKER_SERVER';
+    else if (!GIT_CLIENT_URL) errorStatus = 'MISSING_GIT_CLIENT_URL';
+    else if (!GIT_URL || !GIT_USERNAME || !GIT_PASSWORD) errorStatus = 'MISSING_GIT_SERVER_CONFIG';
+    if (errorStatus) {
+      logger.error(
+        { ...logContext, errorStatus },
+        'received git client request with wronng configuration'
+      );
+      return emit({
+        status: 500,
```
pkey, line 47: optional: bad request would be better?

client-templates/azure-repos/.env.sample
```diff
@@ -29,3 +29,13 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
...
 BROKER_CLIENT_VALIDATION_URL="https://$AZURE_REPOS_HOST/$AZURE_REPOS_ORG/_apis/git/repositories"
+
+# the url of your snyk git client (including scheme and port)
+# GIT_CLIENT_URL=https://<snyk-git-client-host>:<snyk-git-client-port>
+
+# the host where the git server resides
+GIT_URL=$AZURE_REPOS_HOST
+
+# git credentials for cloning repos
+GIT_USERNAME=PAT
```
pkey, line 40: question: what is pat? 🤔

lib/relay-git.js
neonnoon: > the idea here is to be a bit loose on the request and tolerant with the format to leave a bit of space for further improvements

I see your point! 👍 Although, we can only tolerant on input as long as we can make the right assumptions without breaking things. To this point this is still fine I think.

lib/relay-git.js
shinin-light: It's not a bad request, but a wrong internal configuration. The request might be fine, but the data to inject is not there.

lib/relay-git.js
shinin-light: On a counter argument, different SCM may use different `schema` to access their git server (e.g. `ftp://`) and we could break things by injecting a default schema. Also, with no verification on the `gitURI` format, this tolerance may easily backfire. In any case, we are owning the gitURI format in the requests, so we can sync Broker and Registry accordingly.

pkey:
Closing the pull request as we have decided on the new approach. The description and everything else is still relevant so I will be referencing this in the new pull request.

client-templates/azure-repos/.env.sample
carlos-snyk: Personal Access Token

PR #329, opened 2021-06-16
Feat: container registry agent support for DigitalOcean CR

jurglic:
- [x] Ready for review

#### What does this PR do?

This is continuation of #328.

It adds support for newly added DigitalOcean CR, which has specific credentials format.

Additionally, order of args in test assertions is fixed in the last commit.


snaftaly (approved):
👏

snaftaly (approved):

PR #330, opened 2021-06-17
feat: add redirect rule to git client

pkey:
- [x] Ready for review

#### What does this PR do?
As part of the efforts to allow on-prem customers to have Snyk Code analysis results, we've built a support service that can be run alongside the Broker Client to clone the repositories and upload the code to our servers to be analyzed. In order to support this solution, we've updated the Broker Client to be aware of this new architecture and redirect all the requests to a specific endpoint towards the new snyk git client.

The expected behaviour at the moment is:
- a new `GIT_CLIENT_URL` variable is introduced in the `.env` file to be specified by the customer during the deployment, alongside three support variables `GIT_URL`, `GIT_USERNAME`, `GIT_PASSWORD` that are just used to capture under a common name different SCM-related variables that already present.
- every request towards `/snykgit/*` is redirected to the relative `snykgit/*` path on the git client.
- the creds in the body of the request and part of the girURI will be replaced by the above mentioned environment variables (this is  not done in this PR)
- no change is required for the Snyk Broker Server, but beta customers will need to update their broker client version and configuration together with deploying the new git client to benefit from on-prem code analyses.

Possible future scenarios:
- we expect other products and teams to start using the snyk git client in the future after the communication protocol is established. For this reasons we want the broker changes to be simple and allow us a variety of approaches end-to-end. 

There are some design choices that may be useful to review for maintainability and longevity of the product:
- use of a path specification to specify the direction of the forwarding (scm vs git-client). This could also be achieved through a custom header, if that's advisable.
- use of env variables to have a common scm-agnostic name reference to info that is already present elsewhere in the config (e.g. GITHUB_TOKEN => GIT_PASSWORD)
- the only env variable that has been added to the dockerfiles examples is the `GIT_CLIENT_URL`. 

- We did test the code locally and added some tests for the new functionality, however the end-to-end interaction with real SCMs has not being tested at all (all the e2e tests have been performed for a brokered github.com integration).

We had multiple design iterations for possible on-prem analysis flow, all of which included some changes to the broker client.

(https://docs.google.com/presentation/d/16U72rCxbrTO_lCBqU6FeNDwpzkPvwsXvJGncjF8PmtM/edit?usp=sharing) or chat with us at [#feat-code-access-2021](https://snyk.slack.com/archives/C01T8J66QMS)

### DIFF
--- a/client-templates/azure-repos/.env.sample
+++ b/client-templates/azure-repos/.env.sample
@@ -29,3 +29,13 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
 
 # Azure validation url, checked by broker client systemcheck endpoint
 BROKER_CLIENT_VALIDATION_URL="https://$AZURE_REPOS_HOST/$AZURE_REPOS_ORG/_apis/git/repositories"
+
+# the host where the git server resides
+GIT_URL=$AZURE_REPOS_HOST
+
+# git credentials for cloning repos
+GIT_USERNAME=PAT
+GIT_PASSWORD=$AZURE_REPOS_TOKEN
+
+# the url of your snyk git client (including scheme and port).
+# GIT_CLIENT_URL=https://<snyk-git-client-host>:<snyk-git-client-port>
--- a/client-templates/gitlab/.env.sample
+++ b/client-templates/gitlab/.env.sample
@@ -23,3 +23,13 @@ ACCEPT=accept.json
 
 # The path for the broker's internal healthcheck URL. Must start with a '/'.
 BROKER_HEALTHCHECK_PATH=/healthcheck
+
+# the host where the git server resides
+GIT_URL=$GITLAB
+
+# git credentials for cloning repos
+GIT_USERNAME=oauth2
+GIT_PASSWORD=$GITLAB_TOKEN
+
+# the url of your snyk git client (including scheme and port).
+# GIT_CLIENT_URL=https://<snyk-git-client-host>:<snyk-git-client-port>
--- a/dockerfiles/azure-repos/Dockerfile
+++ b/dockerfiles/azure-repos/Dockerfile
@@ -10,15 +10,15 @@ RUN apt-get update && apt-get install -y ca-certificates
 
 ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
 
-ENV PATH=$PATH:/home/node/.npm-global/bin 
+ENV PATH=$PATH:/home/node/.npm-global/bin
 
-RUN npm install --global snyk-broker 
+RUN npm install --global snyk-broker
 
 
 
 FROM snyk/ubuntu
 
-ENV PATH=$PATH:/home/node/.npm-global/bin 
+ENV PATH=$PATH:/home/node/.npm-global/bin
 
 COPY --from=base /home/node/.npm-global /home/node/.npm-global
 
@@ -26,7 +26,7 @@ COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificat
 
 # Don't run as root
 WORKDIR /home/node
-USER node 
+USER node
 
 # Generate default accept filter
 RUN broker init azure-repos
--- a/lib/log.js
+++ b/lib/log.js
@@ -62,6 +62,18 @@ function sanitise(raw) {
     raw = sanitiseConfigVariable(raw, 'CR_AGENT_URL');
   }
 
+  if (config.GIT_USERNAME) {
+    raw = sanitiseConfigVariable(raw, 'GIT_USERNAME');
+  }
+
+  if (config.GIT_PASSWORD) {
+    raw = sanitiseConfigVariable(raw, 'GIT_PASSWORD');
+  }
+
+  if (config.GIT_CLIENT_URL) {
+    raw = sanitiseConfigVariable(raw, 'GIT_CLIENT_URL');
+  }
+
   return raw;
 }
 
--- a/test/functional/server-client.test.js
+++ b/test/functional/server-client.test.js
@@ -42,13 +42,17 @@ test('proxy requests originating from behind the broker server', (t) => {
   process.env.ORIGIN_PORT = echoServerPort;
   process.env.USERNAME = 'user@email.com';
   process.env.PASSWORD = 'aB}#/:%40*1';
+  process.env.GIT_CLIENT_URL = `http://localhost:${echoServerPort}`;
+  process.env.GIT_URL = process.env.GITHUB;
+  process.env.GIT_USERNAME = process.env.USERNAME;
+  process.env.GIT_PASSWORD = process.env.PASSWORD;
   const client = app.main({ port: port() });
 
   // wait for the client to successfully connect to the server and identify itself
   server.io.on('connection', (socket) => {
     socket.on('identify', (clientData) => {
       const token = clientData.token;
-      t.plan(26);
+      t.plan(30);
 
       t.test('identification', (t) => {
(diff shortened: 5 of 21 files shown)


client-templates/gitlab/.env.sample
```diff
@@ -23,3 +23,13 @@ ACCEPT=accept.json
 
 # The path for the broker's internal healthcheck URL. Must start with a '/'.
 BROKER_HEALTHCHECK_PATH=/healthcheck
+
+# the host where the git server resides
+GIT_URL=$GITLAB
+
+# git credentials for cloning repos
+GIT_USERNAME=oauth2
```
ohad2712, line 31: Where did you get those usernames?

client-templates/azure-repos/.env.sample
```diff
@@ -29,3 +29,13 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
 
 # Azure validation url, checked by broker client systemcheck endpoint
 BROKER_CLIENT_VALIDATION_URL="https://$AZURE_REPOS_HOST/$AZURE_REPOS_ORG/_apis/git/repositories"
+
```
ohad2712, line 32: A general question- What about all other integrations that can be brokered? Why are there no `.env` changes for them as well?

dockerfiles/azure-repos/Dockerfile
```diff
@@ -56,6 +56,9 @@ ENV PORT 7341
 # This will be used as the webhook payload URL
 ENV BROKER_CLIENT_URL "https://<broker.client.hostname>:$PORT"
 
+# the url of your snyk git client (including scheme and port)
+# ENV GIT_CLIENT_URL https://<snyk-git-client-host>:<snyk-git-client-port>
```
neonnoon, line 60: question: Why is this commented out? These images are built and deployed to dockerhub, right?

lib/log.js
```diff
@@ -62,6 +62,14 @@ function sanitise(raw) {
     raw = sanitiseConfigVariable(raw, 'CR_AGENT_URL');
```
neonnoon, line 62: question: I noticed that `CR_AGENT_URL` is sanitized here while we didn't add anything for `GIT_CLIENT_URL`. Is this OK?

dockerfiles/azure-repos/Dockerfile
neonnoon: I see that below some Dockerfiles actually also have `ENV BROKER_CLIENT_URL` commented out.
Not that it really makes a difference (those are runtime ENVs which will have be provided with `-e` anyway).
More a question of consistency I guess.

neonnoon (approved):
Thanks, looks good to me. Just some minor questions regarding consistency.

client-templates/azure-repos/.env.sample
pkey: @ohad2712 We are only targeting git based integrations (so no docker, or jira). I believe I changed `.env` for all of them.

client-templates/gitlab/.env.sample
pkey: I think those usernames are standard ones in cases when there is only a user token (or equivalent) used for a specific SCM.

@shinin-light is that right?

lib/log.js
pkey: Good question, we probably should. @shinin-light what do you think?

dockerfiles/azure-repos/Dockerfile
pkey: Yeah, I guess it's commented because this is a non-standard env variables and is only required in certain cases at the moment thus it needs to be passed explicitly.

client-templates/gitlab/.env.sample
shinin-light: Sadly there is no standard across the SCMs for the git protocol. This specific username comes from the gitlab specifications, but it should be tested with a real gitlab environment (which I didn't). I didn't find any reasonably quick way of testing the broker with every single SCM integration we support at the moment, but if there is one we should try it out.

ohad2712: Okay I understand.
Isn't just following how to locally set up a brokered integration, while using these vars will be enough for that purpose?

lib/log.js
shinin-light: Good call, I didn't think URLs would be eligible for sanitization, but if one is let's be consistent and sanitize them all.

shinin-light (dismissed):

dockerfiles/azure-repos/Dockerfile
shinin-light: I honestly didn't know why some env were ending up in the dockerfile while others don't and still others end up in the dockerfile but commented out. @pkey explanation above seems sound enough tho.

lib/log.js
pkey: Ok, will add this.

dockerfiles/azure-repos/Dockerfile
pkey: But does it make sense to have them there? Should we just add this to README.md instead? Customer does not really look at those `Dockerfile`s anyways, he just cares about thing he needs to pass to `docker run`

@ohad2712 @carlos-snyk

client-templates/gitlab/.env.sample
pkey: @ohad2712 You mean testing different SCMs?

ohad2712: Yes.

dockerfiles/azure-repos/Dockerfile
carlos-snyk: I think commented out env vars are useless and don't need to be there. Broker users have the `.env.sample` files already to see what env vars are being used and can change

client-templates/gitlab/.env.sample
pkey: Makes sense. We are working on testing, will get back to you once we gone through all of the SCMs.

test/utils.ts
```diff
@@ -1,5 +1,7 @@
 const compression = require('compression');
 const webserver = require('../lib/webserver');
+const express = require('express')
```
pkey, line 3: comment: modified this to accommodate the fact that our requests to the git client have a `snykgit` prefix. My only thought was that the tests might be more generic (instead of using `snykgit` we could have `special-path` for example)

pkey:
I've also ran `npm run format` on everything and some files (`Dockerfiles` mainly) changed.

shinin-light (dismissed):

neonnoon (approved):

test/functional/server-client.test.js
```diff
@@ -406,6 +410,49 @@ test('proxy requests originating from behind the broker server', (t) => {
...
+        const url = `http://localhost:${serverPort}/broker/${token}/snykgit/echo-body`;
+        const body = { some: { example: 'json' } };
+        request({ url, method: 'post', json: true, body }, (err, res) => {
+          t.equal(res.statusCode, 200, '200 statusCode');
+          t.same(res.body, body, 'body brokered');
+          t.end();
+        });
+      });
+
+      t.test('successfully broker exact bytes of POST body', (t) => {
```
carlos-snyk, line 423: should mention `git client` in this and the other tests related to that flow.

Also I'd change `broker` to `redirects` because it makes more sense to me (feel free to break that convention)

carlos-snyk (approved):

test/functional/server-client.test.js
pkey: @carlos-snyk Great suggestions, let me implement those as a last step! Thanks for review 🙏🏻

carlos-snyk (approved):

PR #334, opened 2021-07-01
feat: Log maskedToken on every server request

jurglic:
- [x] Ready for review

#### What does this PR do?

This PR makes some logging improvements on the broker-server:
* masked token is included in every request
* header.X-Broker-Token is removed

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -8,6 +8,7 @@ const { replace, replaceUrlPartialChunk } = require('./replace-vars');
 const tryJSONParse = require('./try-json-parse');
 const logger = require('./log');
 const version = require('./version');
+const { maskToken } = require('./token');
 const stream = require('stream');
 const NodeCache = require('node-cache');
 
@@ -67,6 +68,7 @@ function requestHandler(filterRules) {
       method: req.method,
       headers: req.headers,
       requestId: req.headers['snyk-request-id'] || uuid(),
+      maskedToken: req.maskedToken,
     };
 
     logger.debug(logContext, 'received request over HTTP connection');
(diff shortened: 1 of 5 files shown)


lib/relay.js
```diff
@@ -192,6 +193,7 @@ function responseHandler(filterRules, config, io) {
       headers,
       requestId: headers['snyk-request-id'] || uuid(),
       streamingID,
+      maskedToken: (brokerToken.slice(0, 4) + '-...-' + brokerToken.slice(-4)),
```
carlos-snyk, line 196: feels like the fn inputs [above](https://github.com/snyk/broker/blob/e92bd4d13398e0c00476b9ba9059d271b45e9ad8/lib/relay.js#L187) are coming from the `req`, have you checked whether you can get the `maskedToken` from there?

carlos-snyk (commented):
Looks great

lib/relay.js
jurglic: As far as I can see request is [being emitted on the client](https://github.com/snyk/broker/blob/e92bd4d13398e0c00476b9ba9059d271b45e9ad8/lib/client/index.js#L77), which means request event signature would need to be updated, and hence the client... and afaik it wouldn't change much, as token isn't sent or used in every request, just upon establishing connection with broker... so it's the same if I use the `brokerToken` when initial request handler is initialized.

lib/relay.js
```diff
@@ -396,3 +400,10 @@ function logError(logContext, error) {
     'error while sending websocket request over HTTP connection',
   );
 }
+
+function sanitizeHeaders(headers) {
```
carlos-snyk, line 404: have you seen the existing `sanitiseHeaders` function? Probably should do it there so that it applies everywhere

jurglic: No I haven't. Great call, will amend it!

carlos-snyk (approved):

PR #340, opened 2021-07-06
fix: mitigate engine.io DoS vulnerability by reducing HTTP buffer size

Kirill89:
- [x] Ready for review

#### What does this PR do?

We use `engine.io` < 4.0.0 which is [vulnerable to DoS](https://app.snyk.io/vuln/SNYK-JS-ENGINEIO-1056749). To mitigate the vulnerability we reduce HTTP buffer side down to 1mb as it is done in [the official fix commit](https://github.com/socketio/engine.io/commit/734f9d1268840722c41219e69eb58318e0b2ac6b).

__Note:__ it's okay to keep this change even after updating `engine.io` to >= 4.0.0 because it should not affect anything.

https://app.snyk.io/vuln/SNYK-JS-ENGINEIO-1056749

### DIFF
--- a/lib/server/socket.js
+++ b/lib/server/socket.js
@@ -8,7 +8,13 @@ module.exports = ({ server, filters, config }) => {
   const io = new Primus(server, {
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: '20971520',
+    // engine.io < 4.0.0 is vulnerable to Denial of Service attack. To mitigate
+    // the vulnerability we reduce HTTP buffer size down to 1mb as it is done
+    // in the official fix commit.
+    // For more details:
+    // - https://app.snyk.io/vuln/SNYK-JS-ENGINEIO-1056749
+    // - https://github.com/socketio/engine.io/commit/734f9d1268840722c41219e69eb58318e0b2ac6b
+    maxLength: 1e6,
   });
   io.plugin('emitter', Emitter);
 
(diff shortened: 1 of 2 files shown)


lib/server/socket.js
```diff
@@ -8,7 +8,13 @@ module.exports = ({ server, filters, config }) => {
...
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: '20971520',
+    // engine.io < 4.0.0 is vulnerable to Denial of Service attack. To mitigate
+    // the vulnerability we reduce HTTP buffer size down to 1mb as it is done
+    // in the official fix commit.
+    // For more details:
+    // - https://app.snyk.io/vuln/SNYK-JS-ENGINEIO-1056749
+    // - https://github.com/socketio/engine.io/commit/734f9d1268840722c41219e69eb58318e0b2ac6b
+    maxLength: 1e6,
```
carlos-snyk, line 17: This has an impact on Broker communications, we raised it to 20MB a few weeks ago because payloads bigger than 10MB were being blocked. This is going to make things a lot worse

carlos-snyk (changes requested):
Blocking because this will impact Broker users negatively

lib/server/socket.js
Kirill89: Would you mind to add a comment about it in the code?

carlos-snyk: Adding comment: https://github.com/snyk/broker/pull/341

PR #344, opened 2021-07-26
feat: CRA systemchecks

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?

enable systemcheck for the entire CRA flow: broker client -> CRA ->
container registry. We use a dedicated header for passing the
credentials, which will only work for the systemcheck endpoint of the
CRA.


snaftaly (commented):
LGTM. Are there tests we can add for this flow?

PR #347, opened 2021-08-01
feat: systemcheck for container integration

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?
If a user sets `BROKER_CLIENT_VALIDATION_URL` then we also set
`BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER` with whatever is in
CR_CREDENTIALS. The CRA will process the credentials in the systemcheck
endpoint.

### DIFF
--- a/bin/container-registry-agent/docker-entrypoint.sh
+++ b/bin/container-registry-agent/docker-entrypoint.sh
@@ -20,6 +20,9 @@ if [[ -z "$CR_CREDENTIALS" ]]; then
 
   export CR_CREDENTIALS=$(echo $JSON | base64 -w0)
 fi
+if [[ $BROKER_CLIENT_VALIDATION_URL ]]; then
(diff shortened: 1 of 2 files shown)


snaftaly (dismissed):
Is there some test we can add to check the value of the header?
Also, not sure about it, but maybe worth adding a note in the README about how systemcheck works for the CR broker client.

yaronschwimmer:
> Is there some test we can add to check the value of the header?
> Also, not sure about it, but maybe worth adding a note in the README about how systemcheck works for the CR broker client.

@snaftaly I'll see if there's a way to test this somehow. The README can stay as is, because the only thing that needs to be added is `BROKER_CLIENT_VALIDATION_URL` like in other brokered setups. I will update the public docs with more detail after this is released

snaftaly (approved):
Nice! 👏

PR #353, opened 2021-08-17
fix: ubuntu 20.10->21.04, node 12.22.3->12.22.5

neonnoon:
- [x] Ready for review

#### What does this PR do?

- Upgrades base image Ubuntu version to 21.04
- Upgrades node 12 patch level


neonnoon:
See follow up to address node only for now: #354

PR #355, opened 2021-08-20
feat: comment out default envvars in dockerfiles

JackuB:
To allow broker configuration by mounting a .env file.
If you do that today, those default ENV variables will override it. We should keep them only as comments.


yaronschwimmer (approved):
LGTM

iamtmrobinson (approved):

PR #367, opened 2021-10-04
fix: replace partial chunk

gitphill:
- [X] Ready for review

#### What does this PR do?
Fix bug replacing 'RES_BODY_URL_SUB' in relay.
The function returns an object with newChunk.
This should have been assigned to responseBody instead of the entire
object.

The effect of this was to transform our responses from a string to an object,
which when we attempt to write to the http response it throws 'ERR_INVALID_ARG_TYPE'.

Added 'RES_BODY_URL_SUB' env var to server-client functional test to
assert desired behaviour.

Without the fix to relay, you can now see the functional server-client test fail.

### DIFF
--- a/test/functional/server-client.test.js
+++ b/test/functional/server-client.test.js
@@ -46,6 +46,7 @@ test('proxy requests originating from behind the broker server', (t) => {
   process.env.GIT_URL = process.env.GITHUB;
   process.env.GIT_USERNAME = process.env.USERNAME;
   process.env.GIT_PASSWORD = process.env.PASSWORD;
+  process.env.RES_BODY_URL_SUB = `http://private`;
   const client = app.main({ port: port() });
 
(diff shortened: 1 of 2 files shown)


test/functional/server-client.test.js
```diff
@@ -420,19 +421,22 @@ test('proxy requests originating from behind the broker server', (t) => {
...
+          );
+          const headers = { 'Content-Type': 'application/json' };
+          request({ url, method: 'post', headers, body }, (err, res) => {
+            const responseBody = Buffer.from(res.body);
+            t.equal(res.statusCode, 200, '200 statusCode');
+            t.same(responseBody, body, 'body brokered exactly');
+            t.end();
+          });
+        },
+      );
```
gitphill, line 439: this is just prettier re-formatting

yeforriak (approved):

PR #373, opened 2021-11-18
Fix/query param validation

carlos-snyk:
Note: I ran `npx run format` and that changed a few files... Try `hide whitespace` when reviewing the changes 🙏 

About the PR: query param validation was succeeding when a single query param was valid, even if the rest weren't. This is unexpected behaviour. Has now been changed to force all query params to be present and valid.

About the impact: only the Azure Repos accept.json sample is affected by this, because it's the only one with more than one query param. Affected functionality is downloading files and fetching repo file names. I'm testing those flows now.


lib/token.js
```diff
@@ -3,8 +3,8 @@ module.exports = {
 };
 
 function maskToken(token) {
-  if (!token || token === "") {
-    return "";
+  if (!token || token === '') {
```
pavel-solodilov, line 6: **nitpick**: an empty string is falsy
```suggestion
  if (!token) {
```

pavel-solodilov (approved):

yeforriak (approved):

PR #375, opened 2021-12-02
fix: Change documentation language referencing Bitbucket Server to Bitbucket Data Center.

carwin:
- [x] Ready for review

#### What does this PR do?
This PR is a simple documentation update. 

README.md

N/A

Bitbucket Server is now a
deprecated service and Atlassian is moving to Bitbucket Data Center
which is essentially the second iteration of BB Server.

Broker should continue to work as-is but the documentation could be
confusing for those looking specifically for Bitbucket Data Center
support.

#### Screenshots
N/A

#### Additional questions
N/A


pavel-solodilov (commented):
Could we please keep both? It would be equally as confusing for the people looking for Bitbucket Server. I am quite sure that these two will also have different APIs, so we may end up actually having different implementations for them.

carwin:
As far as I know Bitbucket Server is going away and is replaced by Bitbucket Data Center - I'm sure there are still legacy users though. The API endpoints seem the same to me so far, I've been told you can use Broker with Bitbucket Data Center today.

It might make sense to just have a little note in there about Bitbucket Server being deprecated, but let readers know that the same configuration examples will work for both. What do you think?

pavel-solodilov:
> It might make sense to just have a little note in there about Bitbucket Server being deprecated, but let readers know that the same configuration examples will work for both. What do you think?

Yes, I think that this is good! Sorry for taking so long to come back to you; on-call life has not been kind 🤗

carwin:
Turns out GitHub added support for footnotes last year, which seems like a perfect fit for this kind of use case. How's this?

image
...
image

aarlaud:
Closing this as stalled. Most likely addressed with the docs update.

PR #378, opened 2021-12-14
chore: add instructions to create standalone executable

iamtmrobinson:
- [x] Ready for review

#### What does this PR do?

Adds instructions for packaging the Broker as a standalone executable using pkg

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Known Vulnerabilities](https://snyk.io/test/github/snyk/broker/badge.svg?style=flat-square)](https://snyk.io/test/github/snyk/broker)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -459,33 +457,64 @@ docker run --restart=always \
...
+}
+```
+
+Once you have added this, run `pkg` from the root directory of the project like so:
+
+```
+npx pkg .
+```
+
+This will generate executables for whichever targets you have included in the `package.json` in the `binary-releases` directory.
```
iamtmrobinson, line 517: This is the bit I've actually changed, the rest is auto-formatting changes.

PR #386, opened 2022-02-21
feat: do not expand variables that contain passwords

yeforriak:
Avoid expanding variables that contain passwords since passwords may contain `$` symbols 

- [x] Ready for review

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -13,6 +13,15 @@ function camelify(res) {
   }, {});
 }
 
+function isEnvVarContainingPassword(envVarName) {
+  const enVarsContainingPasswords = [
+    'PASSWORD',
+    'BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER',
+    'BROKER_CLIENT_VALIDATION_BASIC_AUTH',
+  ];
+  return enVarsContainingPasswords.some((v) => envVarName.includes(v));
+}
+
 function expandValue(obj, value) {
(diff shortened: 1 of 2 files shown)


ar-snyk (approved):
no issues found

lib/config.js
```diff
@@ -13,6 +13,15 @@ function camelify(res) {
   }, {});
 }
 
+function isEnvVarContainingPassword(envVarName) {
+  const enVarsContainingPasswords = [
+    'PASSWORD',
+    'BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER',
+    'BROKER_CLIENT_VALIDATION_BASIC_AUTH',
+  ];
```
carlos-snyk, line 21: The Broker code doesn't know anything about SCMs or their env vars. I think it should stay that way

carlos-snyk (changes requested):

lib/config.js
carlos-snyk: Maybe the solution is to have an env var with the property names that contain secrets.

`ENV_VARS_CONTAINING_PASSWORDS="BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER, BROKER_CLIENT_VALIDATION_BASIC_AUTH"`

obenn:
Issue addressed in https://github.com/snyk/broker/pull/406, suggest close if 406 merged

PR #392, opened 2022-04-06
feat: observe broker response size

gitphill:
- [X] Ready for review

#### What does this PR do?
Introduces two new metrics:

* `broker_response_size_bytes`
* `broker_unable_to_size_response_count`

`broker_unable_to_size_response_count` can be removed if we see that all responses are being observed (this metric is empty).

`broker_response_size_bytes` starts at 1kb and goes upto the maximum length 20mb and is labeled with `isStreaming` (boolean).

### DIFF
--- a/lib/relay.js
+++ b/lib/relay.js
@@ -11,6 +11,7 @@ const version = require('./version');
 const { maskToken } = require('./token');
 const stream = require('stream');
 const NodeCache = require('node-cache');
+const metrics = require('./metrics');
 
 module.exports = {
   request: requestHandler,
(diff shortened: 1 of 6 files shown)


lib/relay.js
```diff
@@ -145,6 +156,24 @@ function requestHandler(filterRules) {
...
+              const responseBodyBytes = Buffer.byteLength(
+                responseBodyString,
+                'utf-8',
+              );
+              metrics.observeResponseSize({
+                bytes: responseBodyBytes,
+                isStreaming: false,
+              });
+            } else {
+              metrics.incrementUnableToSizeResponse();
```
jk05, line 175: maybe add a comment here that we don't expect this to be hit?

yeforriak (approved):

PR #393, opened 2022-04-08
feat: count how many tgz file are requested

gitphill:
- [X] Ready for review

#### What does this PR do?
Count how many tgz file are requested.

Also `JSON.stringify` all non-string response types when calculating response sizes.


ipsi:
Heya, I don't think this metric is something that we need - I think that file size is a better metric for us and that's something we're already tracking.

Given that, I'll go ahead and close the issue.

PR #394, opened 2022-04-13
feat: broker executables

carlos-snyk:
This PR introduces `pkg` for creating Broker executables that don't require Nodejs in the machine.

### DIFF
--- a/client-templates/github
+++ b/client-templates/github
@@ -1,1 +1,0 @@
-./github-com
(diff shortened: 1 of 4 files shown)


yeforriak (dismissed):

Jimimaku:
> This PR introduces `pkg` for creating Broker executables that don't require Nodejs in the machine.

PR #399, opened 2022-05-03
fix: correct websocket config types

soniqua:
- [x] Ready for review

#### What does this PR do?
- Ensures that websocket configuration parameters are of the expected type - previously these were set as `string`, which the underlying `primus` library may not correctly respect.

- Compare the types expected by `Primus` (https://github.com/primus/primus#getting-started) against the types specified in both `lib/client/socket.js` and `/lib/client/server.js`.

I've tested this by:
- Executing the project locally without any env vars to observe type of default configuration (logging the `io.options` object)
- Specifying environment variables (`SOCKET_MAX_RESPONSE_LENGTH="200000"`) and ensuring the value is updated in the `io.options` object, and that the type is correct (here, `Int`).

-

#### Screenshots
-

### DIFF
--- a/lib/server/socket.js
+++ b/lib/server/socket.js
@@ -12,11 +12,11 @@ module.exports = ({ server, filters, config }) => {
   const ioConfig = {
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: config.socketMaxResponseLength ?? '20971520', // support up to 20MB in response bodies
+    maxLength: parseInt(config.socketMaxResponseLength) ?? 20971520, // support up to 20MB in response bodies
     transport: { allowEIO3: true },
-    pingInterval: config.socketPingInterval ?? 30000,
-    compression: config.socketUseCompression ?? false
-  }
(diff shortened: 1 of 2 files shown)


lib/server/socket.js
```diff
@@ -12,11 +12,11 @@ module.exports = ({ server, filters, config }) => {
...
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: config.socketMaxResponseLength ?? '20971520', // support up to 20MB in response bodies
+    maxLength: parseInt(config.socketMaxResponseLength) ?? 20971520, // support up to 20MB in response bodies
     transport: { allowEIO3: true },
-    pingInterval: config.socketPingInterval ?? 30000,
-    compression: config.socketUseCompression ?? false
-  }
+    pingInterval: parseInt(config.socketPingInterval) ?? 30000,
+    compression: Boolean(config.socketUseCompression) ?? false,
```
carlos-snyk, line 18:
```suggestion
    compression: Boolean(config.socketUseCompression),
```

carlos-snyk (approved):

PR #401, opened 2022-05-06
feat[NEBULA-547]: switch to sockjs

pkey:
Assumption is that sockjs does polling differently and might handle huge
responses better


pkey:
not relevant

PR #403, opened 2022-05-13
chore: reduce stream cache ttl and add transport log

nick-watson:


lib/relay.js
```diff
@@ -112,6 +109,7 @@ function requestHandler(filterRules) {
         streams.set(streamingID, {
           response: res,
           streamBuffer,
+          streamSize: 0
```
slyg, line 112: Agreed it is better to move this to the cache 👍

slyg (approved):

PR #404, opened 2022-05-18
fix: remove use of ?? for config values

nick-watson:
parseInt(undefined) = NaN
NaN ?? 1000 = NaN

Therefore use || as the fallback operator

### DIFF
--- a/lib/server/socket.js
+++ b/lib/server/socket.js
@@ -12,11 +12,12 @@ module.exports = ({ server, filters, config }) => {
   const ioConfig = {
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: parseInt(config.socketMaxResponseLength) ?? 20971520, // support up to 20MB in response bodies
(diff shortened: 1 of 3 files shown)


slyg (approved):

lib/server/socket.js
```diff
@@ -12,10 +12,10 @@ module.exports = ({ server, filters, config }) => {
   const ioConfig = {
     transformer: 'engine.io',
     parser: 'EJSON',
-    maxLength: parseInt(config.socketMaxResponseLength) ?? 20971520, // support up to 20MB in response bodies
+    maxLength: config.socketMaxResponseLength ?? 20971520, // support up to 20MB in response bodies
```
slyg, line 15: We could possibly have used `parseInt(config.socketMaxResponseLength ?? 20971520)`

slyg (changes requested):
Actually, using the `||` operator is probably safer

slyg (approved):

PR #406, opened 2022-05-25
fix: env var substitution

obenn:
- [x] Ready for review

#### What does this PR do?
Addresses current environment variable substitution logic breaking passwords or any field that contains a $ character

lib/config.js lines 19/20
tests/unit/config.tests.js lines 6/13

enforce BITBUCKET_PASSWORD="abc$123" to be retained in config object after evaluate method

Very important to ensure users don't need to unnecessarily change their passwords and reduce security.


carlos-snyk (approved):

obenn:
Fixed in https://github.com/snyk/broker/pull/408

PR #407, opened 2022-05-26
Update README.md

pkey:
following the dogfooding, I've found that it was not clear where to find broker token and personal access token for GitHub. Hope that adding links will make it more clear for someone who will try to set this up from scratch.


pkey:
@snyk/comet I know you are going through ownership switch but I think this one is easy 💸

neonnoon:
@pkey is this still a thing?

PR #409, opened 2022-05-31
feat: use RH ubi8 as base image for Citibank

ohad2712:


snaftaly (commented):
Overall looks good to me.
Would this change also affect the images we build for snyk/broker in DockerHub?
If so, worth adding a new Dockerfile.ubi8 file for the these ubi specific images. 
Wdyt?

dockerfiles/artifactory/Dockerfile
```diff
@@ -1,31 +1,53 @@
-FROM snyk/ubuntu as base
+# Global argument
+ARG NODEJS_VERSION=16.15.0
```
snaftaly, line 2: minor comment - i'd suggest to put the ARG below the FROM statement

PR #412, opened 2022-06-10
feat: add Nexus 2 client-templates and dockerfile & update Nexus 3 BROKER_CLIENT_VALIDATION_URL

gitphill:
- [x] Ready for review

#### What does this PR do?

- Update `BROKER_CLIENT_VALIDATION_URL` for `nexus` `.env.sample`
- Create Nexus 2 client-templates and dockerfile (`nexus2`). `nexus` will refer to Nexus 3 as before
- Changing accept.json method 'any' on `NEXUS_URL` to `GET` to be more specific on what Snyk requires access to, for `nexus` and `nexus2`
- The [README will be updated for Nexus 3 ](https://github.com/snyk/broker/pull/417) but only merged before GA. Nexus 2 will be going into beta so not for now.

### DIFF
--- a/client-templates/nexus2/.env.sample
+++ b/client-templates/nexus2/.env.sample
@@ -6,7 +6,7 @@ BROKER_TOKEN=<broker-token>
 BASE_NEXUS_URL=https://<username>:<password>@<your.nexus.hostname>
 
 # The URL to your Nexus Repository Manager
-NEXUS_URL=$BASE_NEXUS_URL/repository
+NEXUS_URL=$BASE_NEXUS_URL/nexus/content
 
 # The URL of the Snyk broker server
 BROKER_SERVER_URL=https://broker.snyk.io
@@ -20,14 +20,8 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
 
 # Provide RES_BODY_URL_SUB with the URL of the Nexus without credentials
 # This URL substitution is required for NPM integration
-# RES_BODY_URL_SUB=https://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/nexus/content
 
-# Nexus validation url, checked by broker client systemcheck endpoint
-# users must have 'nx-metrics-all' privilege enabled to access
-# https://support.sonatype.com/hc/en-us/articles/226254487-System-Status-and-Metrics-REST-API
-# NOTE! If NOT using username and password in BASE_NEXUS_URL, because your Nexus server is open on a private network, then use '/service/rest/v1/status' for BROKER_CLIENT_VALIDATION_URL.
-# '/service/rest/v1/status/check' requires an authenticated connection while '/service/rest/v1/status' does not.
-#BROKER_CLIENT_VALIDATION_URL=$BASE_NEXUS_URL/service/rest/v1/status
-BROKER_CLIENT_VALIDATION_URL=$BASE_NEXUS_URL/service/rest/v1/status/check
+BROKER_CLIENT_VALIDATION_URL=$BASE_NEXUS_URL/nexus/service/local/status
 
 BROKER_CLIENT_VALIDATION_JSON_DISABLED=true
--- a/client-templates/nexus2/accept.json.sample
+++ b/client-templates/nexus2/accept.json.sample
@@ -14,12 +14,7 @@
     },
     {
       "method": "GET",
-      "path": "/service/rest/v1/status",
-      "origin": "${BASE_NEXUS_URL}"
-    },
-    {
-      "method": "GET",
-      "path": "/service/rest/v1/status/check",
+      "path": "/nexus/service/local/status",
       "origin": "${BASE_NEXUS_URL}"
     },
     {
--- a/dockerfiles/nexus2/Dockerfile
+++ b/dockerfiles/nexus2/Dockerfile
@@ -27,7 +27,7 @@ WORKDIR /home/node
 USER node
 
 # Generate default accept filter
-RUN broker init nexus
+RUN broker init nexus2
 
 ######################################
 # Custom Broker Client configuration #
@@ -40,13 +40,13 @@ RUN broker init nexus
 
 # The URL to your Nexus Repository Manager
 # NOTA BENE: without `/` character in the end of URL
-# If using basic auth, use https://<username>:<password>@<your.nexus.hostname>/repository
-# else use https://<your.nexus.hostname>/repository
-# ENV NEXUS_URL=http://<username>:<password>@<your.nexus.hostname>/repository
+# If using basic auth, use https://<username>:<password>@<your.nexus.hostname>/nexus/content
+# else use https://<your.nexus.hostname>/nexus/content
+# ENV NEXUS_URL=<username>:<password>@<your.nexus.hostname>/nexus/content
 
 # Provide RES_BODY_URL_SUB with the URL of the nexus without credentials and http protocol
 # This URL substitution is required for NPM integration
-# RES_BODY_URL_SUB=http://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=http://<your.nexus.hostname>/nexus/content
 
 # The port used by the broker client to accept internal connections
 # Default value is 7341


client-templates/nexus2/.env.sample
```diff
@@ -0,0 +1,27 @@
...
+ACCEPT=accept.json
+
+# The path for the broker's internal healthcheck URL. Must start with a '/'.
+BROKER_HEALTHCHECK_PATH=/healthcheck
+
+# Provide RES_BODY_URL_SUB with the URL of the Nexus without credentials and http protocol
+# This URL substitution is required for NPM integration
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/service/local/repositories
+
+BROKER_CLIENT_VALIDATION_URL=$BASE_NEXUS_URL/service/local/status
```
gitphill, line 25: need to check this and add some commentry if its different when using auth vs not using auth

client-templates/nexus3/.env.sample
```diff
@@ -20,10 +20,17 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
...
 # This URL substitution is required for NPM integration
-# RES_BODY_URL_SUB=http://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/repository
 
 # Nexus validation url, checked by broker client systemcheck endpoint
 # readonly users must have 'nx-metrics-all' privilege enabled to access
 # https://support.sonatype.com/hc/en-us/articles/226254487-System-Status-and-Metrics-REST-API
+
+# NOTE! If NOT using username and password in BASE_NEXUS_URL, because your Nexus server is open on a private network,
+# then use '/rest/v1/status' for BROKER_CLIENT_VALIDATION_URL.
```
mika-bar, line 30: use `'/service/rest/v1/status'` for better readability, perhaps? (according to the REST API docs)

client-templates/nexus3/.env.sample
```diff
@@ -20,10 +20,17 @@ BROKER_HEALTHCHECK_PATH=/healthcheck
 
 # Provide RES_BODY_URL_SUB with the URL of the Nexus without credentials and http protocol
 # This URL substitution is required for NPM integration
-# RES_BODY_URL_SUB=http://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/repository
```
mika-bar, line 23: is there a specific reason for this change? does `npm-deps` accommodate that?

client-templates/nexus3/accept.json.sample
```diff
@@ -0,0 +1,26 @@
...
+      "method": "HEAD",
+      "path": "/",
+      "origin": "${BASE_NEXUS_URL}"
+    },
+    {
+      "method": "GET",
+      "path": "/service/rest/v1/status/check",
+      "origin": "${BASE_NEXUS_URL}"
+    },
+    {
```
mika-bar, line 20: should we add on the `README.md` explanation for running in unauthenticated mode (without `/check`), using: 
```
 {
      "method": "GET",
      "path": "/service/rest/v1/status",
      "origin": "${BASE_NEXUS_URL}"
 }

**Review by mika-bar — COMMENTED** [review]
we must emphasize it's Nexus >= v2.5, to avoid users using versions lower than that. can we change the folder name to `nexus2.5`, syntactically wise? 
will need to include details re that, in the `README.md` as well.

File: client-templates/nexus2/.env.sample (hunk 1 above)
Comment (mika-bar) [reply client-templates/nexus2/.env.sample:25]: according to the docs- we can use `'$BASE_NEXUS_URL/internal/healthcheck'`. 
will need to validate if it requires `basicAuth` when QA'ing as well. 
See -  
- https://github.com/sonatype/nexus-oss/blob/nexus-2.11.x/components/nexus-webapp/src/main/java/org/sonatype/nexus/webapp/metrics/MetricsModule.java#L94

- https://blog.sonatype.com/learn-the-nexus-rest-api-automating-sonatype-nexus

File: client-templates/nexus2/accept.json.sample [hunk 5]
```diff
@@ -18,7 +18,7 @@
       "origin": "${BASE_NEXUS_URL}"
     },
     {
-      "method": "any",
+      "method": "GET",
```
mika-bar, line 21:
```
{
  "method": "GET",
  "path":"/internal/healthcheck",
  "origin":"${BASE_NEXUS_URL}"
}
```
and for unauthenticated access we can advise the following path instead: 
`"path":"/service/local/status"`
based on the plugin's REST API in Nexus 2.x, see - 
- https://help.sonatype.com/repomanager2/configuration/plugins-and-the-rest-api
- https://repository.sonatype.org/nexus-restlet1x-plugin/default/docs/path__status.html

client-templates/nexus3/.env.sample
jk05: Good point. Just looking at the comment on line 21 it says `http`

gitphill: yea we discovered with customers during private beta this is https for Nexus. It's copied from Artifactory env, where it's a little strange, the tarball urls start http (not https) - some Artifactory bug.

jk05: I see. So its ok to leave as `https` then and remove that part of the comment on line 21?

client-templates/nexus2/.env.sample
mika-bar: no need to use the `/internal` endpoint since trying to hit `BROKER_CLIENT_VALIDATION_URL=$BASE_NEXUS_URL/nexus/service/local/status` with basic auth in front of the nexus domain also works, and returning 401 if wrong creds being passed.

client-templates/nexus2/accept.json.sample
mika-bar: no need to add, from the same reason above.

dockerfiles/nexus3/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Redefine in derived Dockerfile,    #
+# or provide as runtime args `-e`    #
+######################################
+
+# Your unique Broker identifier
+# ENV BROKER_TOKEN <broker-token>
+
+# The URL to your Nexus Repository Manager
+# NOTA BENE: without `/` character in the end of URL
+# ENV NEXUS_URL=<username>:<password>@<your.nexus.hostname>/repository
```
jk05, line 45: not quite correct - appreciate this was how it was before anyway

client-templates/nexus3/accept.json.sample
mika-bar: we can't add notes to accept.json, do you mean on the `README.md `as well as on the `.env` file? 
and the empty path is for commentary purposes, to emphasize the `/check` diff.

dockerfiles/nexus3/Dockerfile
mika-bar: this is the URL for nexus3. what do you mean?

client-templates/nexus3/accept.json.sample
jk05: Yeah I meant a note somewhere, not on the json. Maybe in the README or we could make it clear in the `.env` file comments instead? Im not sure I understand your comment about the empty path. I was referring to use not being able to do this:
```
 {
      "method": "GET",
      "path": "",
      "origin": "BROKER_CLIENT_VALIDATION_URL"
 }
```

dockerfiles/nexus3/Dockerfile
jk05: its missing `https://` etc. Theres probably other missing stuff in these comments. Not a major issue, can look over it

client-templates/nexus3/accept.json.sample
jk05: Also I meant we shouldnt leave both `/service/rest/v1/status/check` and `/service/rest/v1/status` in here. Its technically wrong.

client-templates/nexus3/README.md
```diff
@@ -0,0 +1,21 @@
+# Nexus 3 accept.json
```
jk05, line 1: Added in a `README` for this, or we can highlight this in the main repo README when we come to do it. Not sure where to put this info.

client-templates/nexus2/.env.sample
```diff
@@ -0,0 +1,27 @@
+# your unique broker identifier
+BROKER_TOKEN=<broker-token>
+
+# The Base URL for your Nexus Repository Manager
+# If not using basic auth this will only be "https://<your.nexus.hostname>"
+BASE_NEXUS_URL=https://<username>:<password>@<your.nexus.hostname>
+
+# The URL to your Nexus Repository Manager
+NEXUS_URL=$BASE_NEXUS_URL/nexus/content
```
jk05, line 9: realised it can work with just this, and then updated the [Nexus Integration document](https://docs.google.com/document/d/110yx59Qm-eN8rthJF1pZYlHbyp-MXVDMkQR3tnOUGGM/edit#) to tell users the correct language setup for groups and repositories

client-templates/nexus/.env.sample
```diff
@@ -18,12 +18,18 @@ ACCEPT=accept.json
...
-# RES_BODY_URL_SUB=http://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/repository
 
 # Nexus validation url, checked by broker client systemcheck endpoint
 # readonly users must have 'nx-metrics-all' privilege enabled to access
 # https://support.sonatype.com/hc/en-us/articles/226254487-System-Status-and-Metrics-REST-API
+
+# NOTE! If NOT using username and password in BASE_NEXUS_URL, because your Nexus server is open on a private network, then use '/service/rest/v1/status' for BROKER_CLIENT_VALIDATION_URL.
+# '/service/rest/v1/status/check' requires an authenticated connection while '/service/rest/v1/status' does not.
+# Need to make sure the accept.json also accounts for this endpoint
```
jk05, line 31: Im going to leave a note / example in the README update PR for GA

dockerfiles/nexus/Dockerfile
```diff
@@ -4,6 +4,8 @@ MAINTAINER Snyk Ltd
 
 USER root
 
+RUN apt update && apt install -y make python gcc
```
jk05, line 7: where did this come from? did you add this @gitphill ?

client-templates/nexus/.env.sample
jk05: or shall we leave both `/service/rest/v1/status` and `/service/rest/v1/status/check` in the `accept.json`?

jk05: @gitphill ^

client-templates/nexus/.env.sample
```diff
@@ -18,12 +18,17 @@ ACCEPT=accept.json
 # The path for the broker's internal healthcheck URL. Must start with a '/'.
 BROKER_HEALTHCHECK_PATH=/healthcheck
 
-# Provide RES_BODY_URL_SUB with the URL of the Nexus without credentials and http protocol
+# Provide RES_BODY_URL_SUB with the URL of the Nexus without credentials
 # This URL substitution is required for NPM integration
-# RES_BODY_URL_SUB=http://<your.nexus.hostname>/repository
+# RES_BODY_URL_SUB=https://<your.nexus.hostname>/repository
```
jk05, line 23: should this be uncommented?

client-templates/nexus/.env.sample
jk05: holding off updating https://github.com/snyk/broker/pull/417 until we come to a conclusion here

client-templates/nexus/.env.sample
gitphill: it's only required for npm/yarn - best to let the customer decide I think

client-templates/nexus/.env.sample
gitphill: is `/service/rest/v1/status/*` ok?

client-templates/nexus2/.env.sample
gitphill: yep makes sense

dockerfiles/nexus/Dockerfile
gitphill: yep I think we need to re-visit this entire Dockerfile, the base image for broker was changed, we should start again from an existing Dockerfile (artifactory?) to make sure we cover everything that disappears / appeared

client-templates/nexus/.env.sample
gitphill: ok, lets list both

PR #421, opened 2022-07-27
fix: set the group to root on the node home folder

enzo-cappa:
- [x] Ready for review

#### What does this PR do?

On Openshift it's required that all files are owned by the root group,
and with all perms set. This is because the images run with a randomly
assigned user id, but part of the root group. See
https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#images-create-guide-openshift_create-images

Only one file was changed, but it's good to take a look at https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#images-create-guide-openshift_create-images

Run latest broker Bitbucket image on Openshift

### DIFF
--- a/dockerfiles/bitbucket-server/Dockerfile
+++ b/dockerfiles/bitbucket-server/Dockerfile
@@ -22,6 +22,12 @@ COPY --from=base /home/node/.npm-global /home/node/.npm-global
 
 COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
+# On Openshift it's required that all files are owned by the root
+# group, and with all perms set. This is because the images run with a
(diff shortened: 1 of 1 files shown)


attriaayush (dismissed):

attriaayush (approved):

ipsi:
This would only fix the issue for the BitBucket Broker Client, leaving all the other images with potential issues. If we're going to fix this, I'd prefer a solution that fixes all the images.

attriaayush (changes requested):
https://github.com/snyk/broker/pull/421#issuecomment-1200921192

enzo-cappa:
Will be handled in a different way.

Jimimaku:
> [CLA assistant check](https://cla-assistant.io/snyk/broker?pullRequest=421) All committers have signed the CLA.

PR #423, opened 2022-08-12
feat[NEBULA-752]: cloning through broker

pkey:
- [ ] Ready for review

#### What does this PR do?

Introduces rules that allow cloning through broker

NOTE: not ready for review as it's still being tested


pkey:
@neonnoon @nick-watson Rules are ready. I wonder whether we should share this with comet before we talk with the product and figure out whether we want these to be baked in?

neonnoon:
Closing this for now, we can pick it up once the clone-through-broker problems are addressed and we have an understanding how to go forward with accept.json.

PR #428, opened 2022-09-01
feat: add ability to create a credential array

ipsi:
- [✅] Ready for review
- [✅] Follows [CONTRIBUTING](https://github.com/snyk/broker/blob/master/.github/CONTRIBUTING.md) rules

#### What does this PR do?
Users can now specify a credential pool by adding "_POOL" to the
environment variable name, and providing a comma-separated list of
credentials as the value, which will get turned into an array
internally. For example, to provide multiple GitHub tokens, you would
do:

GITHUB_TOKEN_POOL=token1, token2, token3, ...

This will then create an array, and anywhere that GITHUB_TOKEN is
referenced (e.g., in accept.json) it will instead pick a token from that
list (picking the first then second then third then back to the
beginning, etc).

It will also create an array of any other variables that reference it,
like BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER - it will create a
BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER_POOL value, which is
used by the systemcheck endpoint.

This includes a breaking change to the /systemcheck endpoint as it now
returns a JSON array rather than an object - this is to accommodate
checking multiple credentials at once.

Start in `config.js`, which has to parse the pools, then `replace-vars.js`, which handles the actual replacing. Then `index.js` for the `/systemcheck` endpoint and `filter.js` has some minor changes applied.

Start the Broker Client with an `_POOL` environment variable for whichever provider you're using. Verify the `/systemcheck` endpoint picks up both credentials. Verify multiple requests work. Add one good and one bad credential, and verify 50% of requests fail.

### DIFF
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -62,12 +62,12 @@ module.exports = ({port = null, config = {}, filters = {}}) => {
     }
     let auths = [];
     let rawCreds = [];
-    if (config.brokerClientValidationAuthorizationHeaderArray) {
-      auths = config.brokerClientValidationAuthorizationHeaderArray;
-      rawCreds = config.brokerClientValidationAuthorizationHeaderArray.map(credsFromHeader);
-    } else if (config.brokerClientValidationBasicAuthArray) {
-      auths = config.brokerClientValidationBasicAuthArray.map(s => `Basic ${Buffer.from(s).toString('base64')}`)
-      rawCreds = config.brokerClientValidationBasicAuthArray
+    if (config.brokerClientValidationAuthorizationHeaderPool) {
+      auths = config.brokerClientValidationAuthorizationHeaderPool;
+      rawCreds = config.brokerClientValidationAuthorizationHeaderPool.map(credsFromHeader);
+    } else if (config.brokerClientValidationBasicAuthPool) {
+      auths = config.brokerClientValidationBasicAuthPool.map(s => `Basic ${Buffer.from(s).toString('base64')}`)
+      rawCreds = config.brokerClientValidationBasicAuthPool
     } else if (config.brokerClientValidationAuthorizationHeader) {
       auths.push(config.brokerClientValidationAuthorizationHeader);
       rawCreds.push(config.brokerClientValidationAuthorizationHeader);
--- a/lib/client/index.js
+++ b/lib/client/index.js
@@ -53,15 +53,28 @@ module.exports = ({port = null, config = {}, filters = {}}) => {
     const isJsonResponse = !config.brokerClientValidationJsonDisabled;
 
     // set auth header according to config
+    function credsFromHeader(s) {
+      if (s.indexOf(" ") >= 0) {
+        return s.substring(s.indexOf(" ") + 1);
+      } else {
+        return s;
+      }
+    }
     let auths = [];
+    let rawCreds = [];
     if (config.brokerClientValidationAuthorizationHeaderArray) {
       auths = config.brokerClientValidationAuthorizationHeaderArray;
+      rawCreds = config.brokerClientValidationAuthorizationHeaderArray.map(credsFromHeader);
     } else if (config.brokerClientValidationBasicAuthArray) {
       auths = config.brokerClientValidationBasicAuthArray.map(s => `Basic ${Buffer.from(s).toString('base64')}`)
+      rawCreds = config.brokerClientValidationBasicAuthArray
     } else if (config.brokerClientValidationAuthorizationHeader) {
       auths.push(config.brokerClientValidationAuthorizationHeader);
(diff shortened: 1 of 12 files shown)


lib/client/index.js
```diff
@@ -52,74 +52,58 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
-
-    const validationRequestHeaders = {
-      'user-agent': 'Snyk Broker client ' + version,
-    };
-
     // set auth header according to config
-    if (config.brokerClientValidationAuthorizationHeader) {
-      validationRequestHeaders.authorization =
-        config.brokerClientValidationAuthorizationHeader;
+    function credsFromHeader(s) {
```
nicarl, line 56: Is there a reason the `credsFromHeader` function being defined in the scope of the handler for the `/systemcheck` endpoint?
Extracting it would make the handler less bloated.

lib/client/index.js
```diff
@@ -52,74 +52,58 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+      if (s.indexOf(" ") >= 0) {
+        return s.substring(s.indexOf(" ") + 1);
+      } else {
+        return s;
+      }
+    }
+    let auths = [];
```
nicarl, line 63: Do you think it makes sense to extract line 64-77 to a function?

lib/client/index.js
```diff
@@ -52,74 +52,58 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
-          data.ok = false;
-          data.error =
-            responseStatusCode === 401 || responseStatusCode === 403
-              ? 'Failed due to invalid credentials'
-              : 'Status code is not 2xx';
-
-          logger.error(data, response && response.body, 'Systemcheck failed');
-          return res.status(500).json(data);
-        }
+    const rv = [];
```
nicarl, line 81: `rv` = request validated?
Maybe there is more verbose argument name

package.json
```diff
@@ -73,6 +73,7 @@
     "prom-client": "^11.5.3",
     "qs": "^6.9.1",
     "request": "^2.88.1",
+    "request-promise-native": "^1.0.9",
```
nicarl, line 76: Is there a reason you added this particular dependency? I checked on npm and it is deprecated.
We can use another library: got, axios or node-fetch are popular choices.

lib/client/index.js
```diff
@@ -146,3 +130,72 @@ module.exports = ({ port = null, config = {}, filters = {} }) => {
...
+  const validationRequestHeaders = {
+    'user-agent': 'Snyk Broker client ' + version,
+  };
+  if (auth) {
+    validationRequestHeaders.authorization = auth;
+  }
+
+  let errorOccurred = true;
+  // This was originally `request`, but `await` is a lot easier to understand than nested callback hell.
+  await rp(
```
nicarl, line 152: See related comment on `package.json`

package.json
ipsi: It's the promises version of the `request` library, which is also deprecated, but at least it means we don't have two different HTTP libraries in there.

lib/client/index.js
ipsi: `returnValue`, and yeah, you're right.

lib/client/index.js
ipsi: If I'm extracting the bit below to a function (which I am), then it definitely does.

lib/client/index.js
ipsi: Yes, I think it does.

package.json
ipsi: Or rather, it's a promises + async-compatible wrapper around `request`

lib/client/index.js
```diff
@@ -53,56 +53,34 @@ module.exports = ({port = null, config = {}, filters = {}}) => {
...
-      rawCreds = config.brokerClientValidationBasicAuthPool
-    } else if (config.brokerClientValidationAuthorizationHeader) {
-      auths.push(config.brokerClientValidationAuthorizationHeader);
-      rawCreds.push(config.brokerClientValidationAuthorizationHeader);
-      rawCreds = rawCreds.map(credsFromHeader);
-    } else if (config.brokerClientValidationBasicAuth) {
-      auths.push(`Basic ${Buffer.from(config.brokerClientValidationBasicAuth).toString('base64')}`)
-      rawCreds.push(config.brokerClientValidationBasicAuth);
-    }
+    let {auths, rawCreds} = loadCredentialsFromConfig(config);
```
nicarl, line 56: I think those can be `const`

package.json
nicarl: Fair point. Replacing `request` and the wrapper is out of scope for this change.

nicarl (approved):

PR #445, opened 2022-10-19
feat: post responses to Broker Server instead of sending over WebSocket

ipsi:
One of the major limitations of the Broker at the moment is that it cannot handle responses larger than 20MB. This is a somewhat arbitrarily-chosen limit, but it deals with a real problem: large responses fill up in-memory buffers on the Broker Client faster than the Broker Server can read it. This causes the WebSocket ping/pong frames to get "stuck" behind large amounts of data, which in turn means that the Broker Client *thinks* it hasn't received a response, that the WebSocket connection is therefore dead, and it closes it to open a new one.

This commit fixes that by re-using some "streaming" functionality previously provided for handling Code Agent responses, but separating it completely from the WebSocket connection by having the Broker Client make a "POST" request to the Broker Server.

All requests will now use the streaming functionality, not just requests marked as "streaming" in accept.json (that field is now a bit pointless)

The Broker Server and Client should be compatible with older versions of each other, in the event that someone starts using the newer client before the Server has been deployed.

This provides a couple of advantages:

* We can control back-pressure easily, so that if the in-memory buffer fills up we can pause reading from the SCM, for example, until the buffer has been flushed.
* Even if there are issues with the back-pressure, the WebSocket connection will still be relatively responsive as the ping/pong frames will never be "stuck"
* If the WebSocket connection *is* closed for some reason, any responses will still be processed as they are not dependent on the WebSocket.

The major changes here are:
* Add a new endpoint to server/index.js
* Add support for handling POSTing to relay.js
* Add actual POSTing code in stream-posts.js
* Minimal changes to tests - all existing tests passed without issue, except for the one test that was checking non-streaming metrics, which are no longer done if client and server both support POSTing.
* Update webserver.js to not parse the bodies for a specific content type (otherwise we get a full buffer instead of a stream).
* Add support in filters.js for completely blocking requests if the client doesn't have a certain capability.

- [ ] Ready for review


lib/client/socket.js
```diff
@@ -92,8 +76,36 @@ module.exports = ({ url, token, filters, config, identifyingMetadata }) => {
...
+    // null, undefined, empty, etc.
+    logger.error({ url }, 'missing broker url');
+    const error = new ReferenceError(
+      'BROKER_SERVER_URL is required to connect to the broker server',
+    );
+    error.code = 'MISSING_BROKER_SERVER_URL';
+    throw error;
+  }
+
+  // return createWebSocket(url, token, filters, config, identifyingMetadata);
```
carlos-snyk, line 108:
```suggestion
```

lib/filters.js
```diff
@@ -193,11 +222,16 @@ module.exports = (ruleSource) => {
...
-        break;
+    try {
+      for (const test of tests) {
+        res = test(payload);
+        if (res) {
+          break;
+        }
       }
+    } catch (e) {
+      logger.warn('caught error checking request against rules', { error: e });
```
carlos-snyk, line 233: nitpick
```suggestion
      logger.warn({ error: e }, 'caught error checking request against rules');
```

carlos-snyk (dismissed):

lib/stream-posts.js
```diff
@@ -0,0 +1,126 @@
...
+      .on('error', (e) => {
+        logger.error('received error sending data via POST', e);
+      })
+      .on('response', (r) => {
+        if (r.statusCode !== 200) {
+          logger.error('Received unexpected response uploading data', {
+            statusCode: r.statusCode,
+            statusMessage: r.statusMessage,
+            headers: r.headers,
+            body: r.body?.toString(),
```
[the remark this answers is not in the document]
ipsi: Probably not, but it can also be a `Buffer`, which *does* need to call `toString()` 😕

PR #449, opened 2022-11-03
Fix/make iac and code accept rules enabled via env var

aarlaud:
- [x] Ready for review

#### What does this PR do?

Makes enabling IAC and Code in broker a lot easier, not requiring a custom accept.json file and the file mounting into the container.
Simply adding ACCEPT_CODE=true and/or ACCEPT_IAC=tf,yaml,json (or any combination of these 3 extensions) will now automatically add the rules to the default accept.json for your SCM.

FilterRulesLoader in index.js

Spin up a broker and import.


ipsi (dismissed):

ipsi (dismissed):

lib/filter-rules-loading.js
```diff
@@ -0,0 +1,103 @@
...
+        { accept: extensions },
+        '[MISCONFIGURATION] None of the requested ACCEPT IAC file extensions is compatible',
+      );
+    } else {
+      let template = nestedCopy(
+        filters.private.filter(
+          (entry) =>
+            entry.method == 'GET' && entry.path.includes('requirements'),
+        ),
+      );
```
ipsi, line 33: I'm a little worried that this is going to wind up with some weird, Python-related edge cases, but I don't think the odds of that actually happening are high enough to hold this change back.

ipsi (dismissed):

lib/filter-rules-loading.js
aarlaud: I'm just copying an entry so I don't need to maintain SCM specific things. If there is something else I should copy from it's fine, I just picked one of the file that is less likely to go away any time soon.

neonnoon:
Adding @pkey and @nick-watson to make sure this covers our use case. I think we might need some additional tweaks for auth, at least when I look at #423.

pkey:
Thanks @neonnoon. First of all, very cool change @aarlaud, however, I found some issues.

I ran the code and compared BitBucket results, and found that it doesn't align well with rules in our [previous](https://github.com/snyk/broker/pull/423/files) pull request, which we have tested and verified. In this case, what I got is:

```{
            "//": "allow info refs (for git clone)",
            "method": "GET",
            "path": "*/info/refs*",
            "origin": "https://${BITBUCKET_API}",
            "auth": {
              "scheme": "basic",
              "username": "${BITBUCKET_USERNAME}",
            }
          },
```

, and it should be `"origin": "https://${BITBUCKET}"`. I am pretty sure that cloning wouldn't work when `https://${BITBUCKET_API}"` is used. 

Same issue with GitHub - it renders `GITHUB_API` in the rules, rather than `GITHUB` itself. I've tried testing locally and it didn't work.

test/unit/runtime-rules-hotloading.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    process.env.ACCEPT_IAC = 'yaml';
+    const rules = JSON.parse(
+      loadFixture(path.join('accept', 'ghe-with-code-rules-and-iac-yaml.json')),
+    );
+    const loadedRules = loadFilterRules(
+      'ghe.json',
+      path.join(__dirname, '..', 'fixtures/accept'),
+    );
+
+    expect(loadedRules).toEqual(rules);
```
pkey, line 88: nit: rather than having to create these custom fixtures, you could use Jest [snapshots](https://jestjs.io/docs/snapshot-testing). They would save the results and compare them between test making it potentially easier to spot regressions.

lib/filter-rules-loading.js
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+  }
+  if (process.env.ACCEPT_CODE) {
+    logger.info(
+      { accept: process.env.ACCEPT_CODE },
+      'Injecting Accept rules for Code',
+    );
+    let templateGET = nestedCopy(
+      filters.private.filter((entry) => entry.method == 'GET')[0],
```
pkey, line 63: related to my previous comment, I think that the root cause is this logic as it takes the first `GET` pattern from the rule set and applies it. That pattern is likely taken from request to SCM's API but, in our case, we want to make request to the SCMs host.

Here are the rules, that we verified and that work: https://github.com/snyk/broker/pull/423/files

aarlaud: yes I need to distinguish between the different endpoint in used indeed. Will update, thanks for catching that.

test/unit/runtime-rules-hotloading.test.ts
aarlaud: done

lib/filter-rules-loading.js
aarlaud: should be fixed. Going to review all the snapshots to verify their validity before going further

lib/filter-rules-loading.js
```diff
@@ -0,0 +1,146 @@
...
+          ).length > 0,
+      )[0],
+    );
+
+    templateGET.path = '*/info/refs*';
+    templateGET['//'] = 'allow info refs (for git clone)';
+    templateGET.origin = templateGET.origin
+      .replace('https://${GITHUB_TOKEN}', 'https://pat:${GITHUB_TOKEN}')
+      .replace(
+        'https://${GITLAB_TOKEN}@${GITLAB}',
```
pkey, line 109: peeked at the snapshot, looks like gitlab wasn't added. I see ` "origin": "https://\${GITLAB}",` rather than `https://oauth2:${GITLAB_TOKEN}@${GITLAB}'`.

Everything else looks good.

test/unit/runtime-rules-hotloading.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  test.each(scmRulesToTest)(
+    'Injection of valid CODE rules - Testing %s',
+    (folder) => {
+      process.env.ACCEPT_CODE = 'true';
+      const loadedRules = loadFilterRules(
+        'accept.json.sample',
+        path.join(__dirname, '../..', `client-templates/${folder}`),
+      );
+
+      expect(loadedRules).toMatchSnapshot();
```
pkey, line 76: small thing: since you only really care about `code` rules here, you could explicitly filter those two json entries from `loadedRules` - by, for example, filtering for objects that have `for git clone` string and only then snapshot match. That way, you will have less bloated snapshot file.

lib/filter-rules-loading.js
aarlaud: My bad, noticed the oauth2 for gitlab but missed actually adding the logic for it. Will fix

aarlaud:
> Thanks @neonnoon. First of all, very cool change @aarlaud, however, I found some issues.
> 
> I ran the code and compared BitBucket results, and found that it doesn't align well with rules in our [previous](https://github.com/snyk/broker/pull/423/files) pull request, which we have tested and verified. In this case, what I got is:
> 
> ```
>             "//": "allow info refs (for git clone)",
>             "method": "GET",
>             "path": "*/info/refs*",
>             "origin": "https://${BITBUCKET_API}",
>             "auth": {
>               "scheme": "basic",
>               "username": "${BITBUCKET_USERNAME}",
>               "password": "${BITBUCKET_PASSWORD}"
>             }
>           },
> ```
> 
> , and it should be `"origin": "https://${BITBUCKET}"`. I am pretty sure that cloning wouldn't work when `https://${BITBUCKET_API}"` is used.
> 
> Same issue with GitHub - it renders `GITHUB_API` in the rules, rather than `GITHUB` itself. I've tried testing locally and it didn't work.

Thought info ref was an API call, but guess I'm wrong. Will fix

test/unit/runtime-rules-hotloading.test.ts
aarlaud: Doing the whole thing makes me feel better than there is no side effects on the rest of the accept files. Since copy vs nestedcopy gets a bit tricky, I'd rather being comprehensive even at the cost of bloating snapshots

pkey: makes sense

lib/filter-rules-loading.js
aarlaud: fixed the gitlab case. Though a bit strange, the gitlab_token env var doesn't seem in use so much till now...

ipsi (approved):

PR #461, opened 2022-12-07
Add Broker Dispatcher to Server

ipsi:
- [x] Ready for review

#### What does this PR do?
When the server starts up, it should notify the Dispatcher that a new process is available to handle requests, and when it shuts down it should notify it that it is no longer available.

The Server will also notify the Dispatcher when a new connection is made, and when that connection is closed.

This also includes a TypeScript upgrade since some of our dependencies now depend on later versions of TypeScript.

`dispatcher.js` contains the primary logic, with calls made from `server/index.js` and `server/socket.js`.

### DIFF
--- a/lib/dispatcher.js
+++ b/lib/dispatcher.js
@@ -85,7 +85,7 @@ class DispatcherClient {
             error: e,
             dispatcherUrl: this.#url,
             dispatcherVersion: this.#version,
-            id: this.#id,
+            serverId: this.#id,
           },
           'received error communicating with Dispatcher',
         );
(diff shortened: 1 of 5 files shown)


lib/dispatcher.js
```diff
@@ -0,0 +1,187 @@
...
+        if (cb) cb();
+      });
+  }
+}
+
+function hash(token) {
+  const shasum = crypto.createHash('sha256');
+  shasum.update(token);
+  return shasum.digest('hex');
+}
```
pavel-snyk, line 130: @ipsi, it makes sense to export this function, so I can re-use it as well on the client side.

pavel-snyk (dismissed):
LGTM.

Could we test dispatcher logic, so the flow is covered?

aarlaud (approved):

PR #462, opened 2022-12-12
feat: broker dispatcher client [ED-100]

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR adds the ability to use broker-dispatcher feature on the client side. The functionality is equal to the tested changes in `dev` branch.

Additionally we had to convert old TAP tests to JEST (that's why PR is so big). Some tests are skipped and has to be updated.

Under `lib/client/dispatcher` is all logic related to the communication with Dispatcher API.

- start broker client with `BROKER_HA_MODE_ENABLED=true` envvar.


lib/client/dispatcher/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      });
+    } catch (err) {
+      const timeout = 2 ** attempt * 100;
+      logger.warn(
+        { attempt, timeout },
+        `waiting for ${timeout}ms before next Broker Dispatcher API call`,
+      );
+      Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, timeout);
+    }
+  }
```
ipsi, line 46: Is there a reason we have manual retries here _as well as_ automatic Axios retries? Wouldn't this result in up to 90 retries?

PR #473, opened 2023-01-10
fix: enhance accept rule for GH code snippets

pavel-snyk:
This PR fixes issue with blocking request for code snippets in GitHub.

- [x] Ready for review

#### What does this PR do?

This PR fixes the issue with preview of code snippets and allow to list files in the root of repository and in the nested folders.

The current rule for GitHub code snippets doesn't contain `*` so the regex will work for files in the root only.

`/repos/:name/:repo/contents/:path` -> `/repos/:name/:repo/contents/:path*`


aarlaud:
closing this as unnecessary. Need to beware of encodings specifically. Non encoded uris will trip this up.

PR #484, opened 2023-02-01
fix: actually log resp body when streaming as well

aarlaud:
- [x] Ready for review

#### What does this PR do?

Adds body logging also when the response is being streamed.

Requires the same env vars (LOG_LEVEL debug and LOG_ENABLE_BODY).

### DIFF
--- a/lib/stream-posts.js
+++ b/lib/stream-posts.js
@@ -109,6 +109,7 @@ function pipeStream(
       buffer = initStream(config, brokerToken, streamingID, ioData, serverId);
     })
     .on('data', (chunk) => {
+      logResponse(logContext, 'undefined', {body: chunk.toString()}, config);
       logger.trace(`writing data of length ${chunk.length} to buffer`);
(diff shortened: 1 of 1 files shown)


lib/stream-posts.js
```diff
@@ -109,6 +109,7 @@ function pipeStream(
       buffer = initStream(config, brokerToken, streamingID, ioData, serverId);
     })
     .on('data', (chunk) => {
+      logResponse(logContext, '(streaming)', {body: chunk.toString()}, config);
```
ipsi, line 112: This will split large bodies across multiple log entries, and might cause problems if a binary file (e.g., a ZIP file) is logged (though that should be pretty rare for anything that's not Artifactory or Nexus). Also not thrilled about `status` being `(streaming)` - I'd be inclined to set it to `undefined` and just not include it, but otherwise approved.

ipsi (dismissed):

ipsi (approved):

PR #487, opened 2023-02-03
feat: add support for kubernetes api

punkle:
- [x] Ready for review

#### What does this PR do?

This PR adds support for kubernetes api as a broker client.

- it creates a dockerfile for the broker client
- it creates example configuration
- everything defaults to work in a kube environment to use the pods service account token.

You will need access to the api of a kubernetes cluster and a service account token for that api.

Test 1:
Create the docker container. Then run it locally with the environment variables to run it against a kubernetes api "CLUSTER_URL" and "K8S_SERVICE_ACCOUNT_TOKEN". You may need to also mount a ca file with the correct authority for the cluster. Then confirm that the api service is accessable via the client.

Test 2:
Deploy the container to a kubernetes pod with the default settings. Assign the pod a service account and permissions to access the kube api. Then confirm that the api service is accessable via the client.

```
docker build . -f dockerfiles/kubernetes/Dockerfile -t snyk/broker:kubernetes
```

Running against an external kubernetes api:

```
docker run --restart=always \
           -p 8000:8000 \
           -e BROKER_TOKEN=kube-api \
           -e BROKER_SERVER_URL=https://my.broker.server \
           -e CLUSTER_URL=https://kubernetes.api.somewhere \
           -e K8S_SERVICE_ACCOUNT_TOKEN=service-account-token \
       snyk/broker:kubernetes
```

Running in a kubernetes pod:

```
docker run --restart=always \
           -p 8000:8000 \
           -e BROKER_TOKEN=kube-api \
           -e BROKER_SERVER_URL=http://my.broker.server \
       snyk/broker:kubernetes
```

### DIFF
--- a/dockerfiles/kubernetes/Dockerfile
+++ b/dockerfiles/kubernetes/Dockerfile
@@ -1,0 +1,38 @@
+FROM snyk/ubuntu as base
+
+MAINTAINER Snyk Ltd
+
+USER root
+
+RUN apt update && apt install -y make python gcc
+RUN apt-get update && apt-get install -y ca-certificates
+
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin
+
+RUN npm install --global snyk-broker
+
+
+FROM snyk/ubuntu
+
+ENV PATH=$PATH:/home/node/.npm-global/bin
+
+COPY --from=base /home/node/.npm-global /home/node/.npm-global
+
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+# Don't run as root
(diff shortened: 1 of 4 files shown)


dockerfiles/kubernetes/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=base /home/node/.npm-global /home/node/.npm-global
+
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+# Don't run as root
+WORKDIR /home/node
+USER node
+
+# Generate default accept filter
+# RUN broker init kubernetes
```
punkle, line 29: I could not get this line to work so I replaced it with the following two lines.

It didnt work despite the fact that I created the template sample files.

PR #490, opened 2023-02-06
feat: preflight checks for client [ED-161]

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR adds preflight checks to the Broker client. Preflight checks will be executed by client start and can be accessed via `/health/checks` and `/health/checks/:checkId` endpoints.

Configuration variable for enabling/disabling preflight checks: `PREFLIGHT_CHECKS_ENABLED`

Implementation of `CheckStore` and `CheckService` interfaces are responsible for holding and executing checks. At the moment we have `PreflightCheckStore` as in-memory storage for HTTP checks and `HttpCheckService` to run HTTP requests.

#### Additional information
To speed up test execution locally `npm run test` task is split into:
- `npm run test:unit` -> unit tests
- `npm run test:functional` -> functional tests


lib/client/checks/api/checks-handler.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+      res.status(200).json(checkResults);
+    } catch (error) {
+      logger.error({ error }, 'Error executing http checks');
+      res.status(500).json({ error: error });
+    }
+  };
+};
+
+export const handleCheckIdsRoutes = (
```
aarlaud, line 29: can't this handler be combined with the one above and you just check for the param existence to understand this is a specific checkId check vs general check?

lib/client/checks/index.ts
```diff
@@ -0,0 +1,83 @@
+import logger = require('../../log');
+import { Config } from '../config';
+import { CheckStore } from './check-store';
+import { CheckResult } from './types';
+import { HttpCheckService } from './http/http-check-service';
+import { PreflightCheckStore } from './prefilght-check-store';
```
aarlaud, line 6: typo in the file name here

lib/client/checks/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+  return Promise.resolve(preflightCheckResults);
+}
+
+const configurePreflightCheckStore = async (
+  config: Config,
+): Promise<PreflightCheckStore> => {
+  const preflightCheckStore = new PreflightCheckStore();
+
+  await preflightCheckStore.add(createBrokerServerHealthcheck(config));
```
aarlaud, line 57: it'd be nice to load those dynamically/generically so we can add services without having to add them explicitly here, just by loading everything exported in http-checks maybe?

lib/client/checks/index.ts
pavel-snyk: nice catch!

lib/client/checks/index.ts
pavel-snyk: Sounds like a good idea, let's do it in the next iteration to keep the PR scope small.

aarlaud (approved):

lib/client/checks/api/checks-handler.ts
pavel-snyk: moved this sub-task to the next iteration.

PR #491, opened 2023-02-08
fix: add server_id param only if serverId not empty

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
In case that HA mode is not enabled, we don't add `server_id` query param to socket url.

Running Broker client without HA mode should log url without `server_id` query param, e.g.
```
...
broker client is connecting to broker server (url=https://broker.dev.snyk.io, serverId="")
...
```

### DIFF
--- a/lib/client/socket.js
+++ b/lib/client/socket.js
@@ -21,9 +21,11 @@ function createWebSocket(
     pathname: `/primus/${token}`,
   });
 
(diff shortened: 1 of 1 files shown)


aarlaud (approved):

aarlaud:
For the record here, HA mode is not yet live.

PR #502, opened 2023-03-02
fix: use Axios in Server Dispatcher code

ipsi:
* The Broker Server should retry when it receives an error communicating with the Dispatcher (e.g., temporary DNS resolution failure, sidecar error, etc). The easiest solution involved replacing the deprecated Request library with Axios, which handles the retries automatically.
* Add logging to Axios to note when it fails to make a request, in case we get multiple different types of failure
* Reformat code
* Don't clear prometheus metrics, since this removes the custom metrics we add

- [x] Ready for review


lib/server/prometheus-middleware.ts
```diff
@@ -1,9 +1,7 @@
 import * as promBundle from 'express-prom-bundle';
-import * as prom from 'prom-client';
 import { RequestHandler } from 'express';
 
 export const applyPrometheusMiddleware = (): RequestHandler => {
-  prom.register.clear();
```
pavel-snyk, line 6: @ipsi, if we don't clear and call `applyPrometheusMiddleware` twice, will the custom metrics be added only one time?

lib/dispatcher.js
```diff
@@ -74,66 +71,63 @@ class DispatcherClient {
...
-        'Keep-Alive': 'timeout=60, max=10',
-      },
-      body: JSON.stringify(requestBody),
-    })
-      .on('error', (e) => {
+    try {
+      const response = await axiosInstance.request({
+        url,
+        method,
+        data: requestBody ?? JSON.stringify(requestBody),
```
aarlaud, line 84: wouldn't that be stringifying only null or undefined here?

aarlaud: `The nullish coalescing (??) operator is a logical operator that returns its right-hand side operand when its left-hand side operand is [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null) or [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined), and otherwise returns its left-hand side operand.`

ipsi: Whoops, I will fix that up

pavel-snyk (approved):

lib/dispatcher.js
```diff
@@ -74,66 +71,63 @@ class DispatcherClient {
...
+          'Content-Type': 'application/vnd.api+json',
+          Connection: 'Keep-Alive',
+          'Keep-Alive': 'timeout=60, max=10',
+          'Snyk-Request-Id': requestId,
+        },
+      });
+      const statusCode = response.status;
+      const headers = response.headers;
+      const body = response.data;
+      if (statusCode >= 300) {
```
aarlaud, line 95: we're sure that if we are here it's because the 3 retries failed, and not just the first one while the 2 other keep going in a promise or something?

aarlaud (approved):

PR #504, opened 2023-03-03
fix: check BROKER_TOKEN before calling dispatcher

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR does the following:
- check if `BROKER_TOKEN` config variable exists before making an API call to the dispatcher
- improve check for HA mode. All config values getting from snyk-config are strings, so changing type of `BROKER_HA_MODE_ENABLED` from boolean to string.
- format all source files with Prettier (`npm run format` command) for consistency
- remove a non-relevant NPM task `check-tests`: folder structure is changed.

Take a look on `getServerId()` function.

### DIFF
--- a/lib/client/dispatcher/index.ts
+++ b/lib/client/dispatcher/index.ts
@@ -26,9 +26,17 @@ export function highAvailabilityModeEnabled(config: any): boolean {
 
 export async function getServerId(
   config: any,
-  brokerToken: string,
   brokerClientId: string,
 ): Promise<ServerId | null> {
+  if (!config.BROKER_TOKEN) {
+    logger.error({ token: config.BROKER_TOKEN }, 'missing client token');
+    const error = new Error(
+      'BROKER_TOKEN is required to successfully identify itself to the server',
+    );
+    error.name = 'MISSING_BROKER_TOKEN';
+    throw error;
+  }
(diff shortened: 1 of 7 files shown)


aarlaud (approved):

lib/client/dispatcher/index.ts
```diff
@@ -7,21 +7,36 @@ import { ServerId, getServerIdFromDispatcher } from './dispatcher-service';
...
 
 export function highAvailabilityModeEnabled(config: any): boolean {
-  const haConfig = getHAConfig(config);
+  // high availability mode is disabled per default
+  let highAvailabilityModeEnabled = false;
+
+  const highAvailabilityModeEnabledValue = (config as HAConfiguration)
+    .BROKER_HA_MODE_ENABLED;
+
+  if (typeof highAvailabilityModeEnabledValue !== 'undefined') {
```
aarlaud, line 16: I think we should move this to the config file at some point. It's not a dispatcher related code, it's a config matter to understand and convert string<->boolean.

pavel-snyk: @aarlaud, agree. It's on my todo to move all config-related logic (with keys too) into one place.

PR #509, opened 2023-03-09
fix: add a flag to determine if a client has identified or not so we …

ipsi:
…can properly handle cleanup when the connection is closed

Also add an exception monitor and rejected promise handler to log exceptions and promise rejections not otherwise handled.

- [x] Ready for review


pavel-snyk (approved):
LGTM

PR #521, opened 2023-04-11
feat: allow client to describe bitbucket branches by query string

soniqua:
- [x] Ready for review

#### What does this PR do?

- Adds a rule that allows the Bitbucket Broker to describe branches specified by a query string, enabling Snyk Code analysis. This mechanism is used when a branch contains a `/` character.

- Import a sample repository that has a default branch with a `/` in the name. Observe Snyk Code analysis succeeds after this rule has been added.

### DIFF
--- a/client-templates/bitbucket-server/accept.json.sample
+++ b/client-templates/bitbucket-server/accept.json.sample
@@ -910,6 +910,17 @@
       }
     },
     {
+      "//": "Used to get a branch specified by query string",
+      "method": "GET",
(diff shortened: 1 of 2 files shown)


ipsi:
Can you have a look at fixing the failing tests? They're failing because the Jest fixtures need to be updated - see https://jestjs.io/docs/snapshot-testing#updating-snapshots

aarlaud (approved):

PR #526, opened 2023-04-17
fix: restore functional client-server tests [ED-238]

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This is the first part of re-activating functional client-server tests after TAP->Jest migration.

Start with taking a look on `test/setup` directory, there are helper methods to create webserver that mimics 3rd party SCM ("old UtilServer") and broker client and server.

New [detect-port](https://app.snyk.io/advisor/npm-package/detect-port) package (added to devDependencies) is used to find free TCP port and solve issues with failing tests because of the allocated port

All Jest deprecation and warnings are fixed, so we can upgrade Jest to the latest LTS version later.

- run `npm run test:unit` to execute unit tests
- run `npm run test:functional` to execute functional tests


test/functional/healthcheck.test.ts
```diff
@@ -46,27 +38,30 @@ describe('proxy requests originating from behind the broker client', () => {
...
-      server.io.on('close', () => {
+      bs.server.io.on('close', () => {
         resolve();
       });
     });
   });
 
-  it('server healthcheck', async () => {
+  it.skip('server healthcheck', async () => {
     const response = await axios.get(
```
aarlaud, line 63: Why skip?

aarlaud: Looks like this file is not ready, maybe take it out of this PR?

test/functional/client-ca.test.ts
```diff
@@ -13,8 +13,104 @@
...
+      await bs.server.close();
+    }, 100);
+    await new Promise<void>((resolve) => {
+      bs.server.io.on('close', () => {
+        resolve();
+      });
+    });
+  });
+
+  it.skip('get an error trying to connect to a server with unknown CA', async () => {
```
aarlaud, line 95: Still skipping?

test/functional/no-filter.test.ts
```diff
@@ -70,10 +69,8 @@ describe('no filters broker', () => {
     );
 
     expect(response.status).toEqual(401);
-    expect(response.data).not.toBe({
-      message: 'blocked',
-      reason: 'Request does not match any accept rule, blocking HTTP request',
-      url: '/echo-server',
+    expect(response.data).not.toStrictEqual({
```
aarlaud, line 72: Not sure I like this assertion, kinda prefer the one before in principle

test/functional/server-client.test.ts
```diff
@@ -1,572 +1,605 @@
...
+          'content-type': 'application/json',
+        },
+        timeout: 1000,
+        transformResponse: (r) => r,
+        validateStatus: () => true,
+      },
+    );
+
+    expect(response.status).toEqual(200);
+    expect(response.data.length).toEqual(body.length);
```
aarlaud, line 110: Shouldn't we assert on the body itself?

test/functional/no-filter.test.ts
pavel-snyk: Agree, fixed with 5c4112a.

test/functional/server-client.test.ts
pavel-snyk: Nice catch, fixed with ada9ce2

test/functional/healthcheck.test.ts
pavel-snyk: > Why skip?

For some reason the test stop working locally (infinite retry of the broker client).
It makes sense to take it out of the PR and commit when it works. Done with 2aedffc.

test/functional/client-ca.test.ts
pavel-snyk: I take out this file from the PR (request package throws an exception and exits).
Will tackle it with the next PR. Done with c3e8056.

aarlaud (approved):

PR #543, opened 2023-04-26
Disable var substitution

aarlaud:
- [x] Ready for review

#### What does this PR do?

Adds couple of flags (disableBodyVarsSubstitution, disableHeaderVarsSubstitution) to disable variable replacement in some cases, to alleviate security concerns in some implementation with permissive filter rules.

#### How to use these flags?
When running, export disableHeaderVarsSubstitution=true and/or disableHeaderVarsSubstitution=true env vars to disable env var replacement in headers and/or body.


pavel-snyk (approved):
LGTM

PR #559, opened 2023-06-04
refactor(preflight-checks): refine check interface and type

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR refactors and simplifies preflight-checks feature to make it more configurable and easier to understand.
- added `CheckFn` type returning `CheckResult`. This will allow to add not only http checks, but configuration checks (e.g. mutable exclusive or inclusive options)
- added dedicated `HttpCheck` and `HttpExecutor` to handle executing and retry logic
- removed non-necessary check stores and wiring. At the moment this makes the code harder to read and understand. We might reconsider this in the future


lib/client/checks/http/http-executor.ts
```diff
@@ -0,0 +1,90 @@
...
+import { retry } from '../../retry/exponential-backoff';
+import type { AxiosResponse } from 'axios';
+import type { CheckId, CheckResult, CheckStatus } from '../types';
+
+export async function executeHttpRequest(
+  checkOptions: {
+    id: CheckId;
+    name: string;
+  },
+  httpOptions: {
```
aarlaud, line 13: should probably allocate for headers values to be passed there too. I think we'll soon have more elaborate checks that are likely to require auth headers and the likes

lib/client/checks/http/http-executor.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+export function convertAxiosResponseToCheckResult(
+  check: { id: CheckId; name: string; url: string; method: 'GET' | 'POST' },
+  response: AxiosResponse,
+): CheckResult {
+  let status: CheckStatus;
+  if (response.status >= 200 && response.status <= 299) {
+    status = 'passing';
```
aarlaud, line 65: I think this is fine for now but more elaborate conditions will be needed. 2xx response won't allow us to verify some more elaborate checks.

aarlaud (approved):

lib/client/checks/http/http-executor.ts
pavel-snyk: good idea, I'll incorporate it in the next iteration.

lib/client/checks/http/http-executor.ts
pavel-snyk: Agree, let's add default statuses in the next iteration with possibility to override it if needed.

PR #562, opened 2023-06-08
chore: add tracing for tracking disp retries

aarlaud:
- [x] Ready for review

#### What does this PR do?
Lowers timeout value for dispatcher notifications and log retries failure more clearly.

Some very sporadic network communication errors appeared in logs. Attempting to log these more clearly to investigate further.

### DIFF
--- a/lib/axios.ts
+++ b/lib/axios.ts
@@ -3,15 +3,16 @@ const logger = require('./log');
 import axiosRetry from 'axios-retry';
 
 const axiosInstance = axios.create({
-  timeout: 5000,
+  timeout: 2500,
 });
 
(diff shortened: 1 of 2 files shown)


lib/axios.ts
```diff
@@ -3,15 +3,16 @@ const logger = require('./log');
 import axiosRetry from 'axios-retry';
 
 const axiosInstance = axios.create({
-  timeout: 5000,
+  timeout: 1000,
```
pavel-snyk, line 6: @aarlaud, since we use `axiosInstance` everywhere for API calls, is 1000 not low?

pavel-snyk (approved):

PR #571, opened 2023-07-11
feat: add server support for universal client requests

aarlaud:
- [x] Ready for review

#### What does this PR do?

Intercept the connection type from the request url and add it as header to the request going downstream to the broker client.

This doesn't change the current behavior for which requests are still expected over /broker/:token. Instead, /broker/universal prefixed urls will then be expected to be followed by the typeID and broker token. This prefix is then stripped off from the downstream request and passed instead as x-snyk specific header so the broker client can then infer the right configuration to inject for that request.
The universal client spike is visible https://github.com/snyk/broker/compare/master...spike/universal-broker


pavel-snyk (approved):

aarlaud:
Closing in favor of another PR following change in strategy for universal broker support.

PR #576, opened 2023-07-20
feat: add universal broker support

aarlaud:
- [ ] Ready for review

#### What does this PR do?

Adds Universal Broker Support both on server and client side.


config.default.json
```diff
@@ -1,3 +1,200 @@
...
+        "GITHUB_GRAPHQL":"api.github.com",
+
+        "BROKER_CLIENT_VALIDATION_URL":"https://$GITHUB_API/user",
+        "BROKER_CLIENT_VALIDATION_AUTHORIZATION_HEADER":"token $GITHUB_TOKEN",
+        "GIT_URL":"$GITHUB",
+        "GIT_USERNAME":"x-access-token",
+        "GIT_PASSWORD":"$GITHUB_TOKEN"
+    },
+    "GITHUB-ENTERPRISE": {
+
```
pavel-snyk, line 44: @aarlaud, do we need GITHUB variable here, so value substitution works as expected?

lib/client/index.js
```diff
@@ -44,7 +45,12 @@ module.exports = async ({ port = null, config = {}, filters = {} }) => {
 
   if (commitSigningEnabled(config)) {
     const commitSigningRules = commitSigningFilterRules();
-    filters.private?.push(...commitSigningRules);
+    if (config.BROKER_BOOT_MODE == 'universal') {
+      filters['github'].private?.push(...commitSigningRules);
+      filters['github-enterprise'].github.private?.push(...commitSigningRules);
+    } else {
+      filters.private?.push(...commitSigningRules);
```
pavel-snyk, line 52: @aarlaud, since commit signing is very GitHub-centric (and checks GitHub API endpoints), should we leave to `github` and `github-enterprise` only?

config.default.json
pavel-snyk: clarified

lib/client/index.js
pavel-snyk: clarified

pavel-snyk (approved):
LGTM

PR #608, opened 2023-09-04
Chore/refactor codebase

aarlaud:
- [x] Ready for review

#### What does this PR do?
Moves the whole code base to Typescript.

This branch changes have been reflected in the dev/testing-typescripted-version branch which built the github-com-dev, gitlab-dev and nexus-dev images.
Can run this to test whatever.

First step is a code base cleanup and future optimization.


lib/client/config.ts
```diff
@@ -29,4 +29,5 @@ export type Config = BackendAPI &
   BrokerClient &
   BrokerServer &
   CommitSigning &
-  HighAvailabilityMode;
+  HighAvailabilityMode &
+  Record<string, any>;
```
pavel-snyk, line 33: @aarlaud, is this "relaxing" of the config we talked about?

lib/version.ts
```diff
@@ -0,0 +1,2 @@
+import packageJson from '../package.json';
```
pavel-snyk, line 1: 🤔 I'm not sure that the package.json can be imported so with ESM (in case of real version).

test/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["**/*.ts"]
+  "include": ["**/*.ts", "unit/client/scm/github/commit.test.ts", "functional/systemcheck.test.ts"]
```
pavel-snyk, line 3: does `"**/*.ts"` catch everything under `unit/**` and `functional/**`?

pavel-snyk (approved):
🎩 solid, LGTM.
p.s.: non-blocking questions only.

lib/client/config.ts
aarlaud: yes, for now we have to type this generically. I don't think that generic part will go away anytime soon tbh, but we'll add other types that we know about while leaving other values possible. Who knows what new thing we'll need to pass from Node....

test/tsconfig.json
aarlaud: ** is supposed to do that yes

lib/version.ts
aarlaud: a test like this works
```
import packageJson from '../../package.json'

describe('Test version JSON', () => {
    it('test name', () => {
        expect(packageJson.name).toEqual('snyk-broker')
    })
})```

PR #609, opened 2023-09-11
chore: build development images with new base image

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR updates Dockerfile for dev images and refactors CircleCI workflow for it:
- `build-and-push-dev-docker-image` is the CircleCI job for building, scanning and pushing dev images
- building, scanning and pushing images are combined to one workflow to keep builds smaller and faster
- `approve-from-team-broker` approval step is removed because CircleCI doesn't support approval from GH team only


.circleci/config.yml
```diff
@@ -99,6 +99,8 @@ commands:
             export IMAGE_TAG_LATEST=<<parameters.image_tag>>
             [[ "$IMAGE_TAG" =~ .*"base".* ]] && export IMAGE_TAG=<<parameters.image_tag>>
             [[ "$IMAGE_TAG" =~ .*"base".* ]] && export IMAGE_TAG_LATEST=<<parameters.image_tag>>
+            [[ "$IMAGE_TAG" =~ .*"dev".* ]] && export IMAGE_TAG=<<parameters.image_tag>>
+            [[ "$IMAGE_TAG" =~ .*"dev".* ]] && export IMAGE_TAG_LATEST=<<parameters.image_tag>>
```
pavel-snyk, line 103: @antoine-snyk, this catches the case for `dev` images (same with `base` images), so we can have single CircleCI command for building images.

aarlaud: These regexes feel not restrictive enough somehow... they mean any tag containing base or dev right? We don't have tags matching that today, but we might in the future, so we probably should make them a bit more restrictive at some point

aarlaud (approved):

.circleci/config.yml
pavel-snyk: Agree, we should make it more restrictive in case some tags in the future may contain these words.

PR #648, opened 2023-10-18
feat: add insecure downstream mode

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds an insecure downstream mode forcing downstream requests to be made over http instead of https.
To use only in particular cases where there is just no other options.

Today, connecting over http to the downstream system requires bringing a custom accept.json with overriden values, causing friction since you not only need to mount a file, but also deal with the GIT and IAC rules you need to add manually if you want them, not making use of the ACCEPT_IAC or ACCEPT_GIT options smoothing out this entire process.

With this change, you can connect over http while still leveraging the ACCEPT_ flags.

### DIFF
--- a/lib/common/http/utils.ts
+++ b/lib/common/http/utils.ts
@@ -1,0 +1,7 @@
+import { log as logger } from '../../logs/logger';
+export const switchToInsecure = (url: string) => {
+  logger.debug({ url }, 'Forcing insecure url');
+  const urlToSwitch = new URL(url);
+  urlToSwitch.protocol = 'http';
+  return urlToSwitch.toString();
+};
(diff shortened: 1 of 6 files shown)


lib/common/http/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { log as logger } from '../../logs/logger';
+export const switchToInsecure = (url: string) => {
+  logger.debug({ url }, 'Forcing insecure url');
+  const urlToSwitch = new URL(url);
+  urlToSwitch.protocol == 'http';
```
pavel-snyk, line 5: @aarlaud, I think it should be assignment instead of comparison
```
urlToSwitch.protocol = 'http';
```

aarlaud: :facepalm: fixed

pavel-snyk (approved):

PR #670, opened 2023-11-23
fix: add missing default cfg file

aarlaud:
- [x] Ready for review

#### What does this PR do?

Adds missing config file to the image


dockerfiles/base/Dockerfile
```diff
@@ -107,6 +107,7 @@ LABEL org.opencontainers.image.authors="team-broker" \
 
 COPY --from=broker-builder /home/node/.npm-global /home/node/.npm-global
 COPY --from=broker-builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY config.default.json config.default.json
```
pavel-snyk, line 110: @aarlaud, let's use this syntax to be sure that the config is in the right place

```
COPY config.default.json /home/node/config.default.json
```

pavel-snyk (approved):

PR #677, opened 2023-12-07
fix: add stream resp err handlr to handle econnreset [ED-319]

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds an extra error handler for stream response back to broker server to hopefully catch elusive unhandled exception when ECONNRESET occur.

### DIFF
--- a/lib/common/http/downstream-post-stream-to-server.ts
+++ b/lib/common/http/downstream-post-stream-to-server.ts
@@ -52,7 +52,6 @@ class BrokerServerPostResponseHandler {
     this.#serverId = serverId;
     this.#requestId = requestId;
     this.#buffer = new stream.PassThrough({ highWaterMark: 1048576 });
-
     this.#buffer.on('error', (e) =>
       logger.error(
         {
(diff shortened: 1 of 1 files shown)


lib/common/http/downstream-post-stream-to-server.ts
```diff
@@ -52,7 +52,7 @@ class BrokerServerPostResponseHandler {
     this.#serverId = serverId;
     this.#requestId = requestId;
     this.#buffer = new stream.PassThrough({ highWaterMark: 1048576 });
-
+    this.#brokerSrvPostRequestHandler;
```
pavel-snyk, line 55: @aarlaud, we don't instantiate this variable here, do we need this line?

pavel-snyk (dismissed):

lib/common/http/downstream-post-stream-to-server.ts
aarlaud: updated, sorry

pavel-snyk (approved):

PR #702, opened 2024-02-07
feat: add fallback api route for webhook if ws conn down [HYB-448]

aarlaud:
- [x] Ready for review

#### What does this PR do?

Adds a fallback mechanism for webhook events to be relayed via HTTPS to the API in case the WS tunnel is down for any reason.

### DIFF
--- a/lib/client/checks/http/index.ts
+++ b/lib/client/checks/http/index.ts
@@ -1,5 +1,4 @@
 import { executeHttpRequest } from './http-executor';
-import { highAvailabilityModeEnabled } from '../../dispatcher';
 import type { CheckResult, HttpCheck } from '../types';
 import type { Config } from '../../types/config';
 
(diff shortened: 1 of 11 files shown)


lib/client/checks/http/index.ts
```diff
@@ -33,11 +32,10 @@ const restApiStatusCheck = (config: Config): HttpCheck => {
     config.API_BASE_URL || config.BROKER_SERVER_URL
       ? config.BROKER_SERVER_URL.replace('//broker.', '//api.')
       : 'https://api.snyk.io';
-  const enabled = highAvailabilityModeEnabled(config);
   return {
     id: 'rest-api-status',
     name: 'REST API Healthcheck',
-    enabled: enabled,
+    enabled: true,
```
pavel-snyk, line 38: @aarlaud, we want to have this check always enabled, right?

pavel-snyk (approved):
LGTM, only one question with skipping HA mode by enabling `rest-api-status` check (not blocking).

PR #713, opened 2024-02-18
feat: add checkmarks rules

assafenso:
- [ ] Ready for review

#### What does this PR do?
adding Checkmarx rules too apprisk config

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -38,7 +38,7 @@
           },
           "method": "post",
           "headers": {
-            "Content-Type": "application/x-www-form-urlencoded"
+            "x-broker-content-type": "application/x-www-form-urlencoded"
           }
         }
       ],
--- a/config.default.json
+++ b/config.default.json
@@ -25,10 +25,30 @@
     },
     "apprisk": {
       "validations": [
+        {
+          "url": "https://$CHECKMARX/cxrestapi/auth/identity/connect/token",
+          "body": {
+            "BROKER_VAR_SUB": ["username", "password"],
+            "username": "${CHECKMARX_USERNAME}",
+            "password": "{CHECKMARX_PASSWORD}",
+            "grant_type": "password",
+            "scope": "sast_rest_api",
+            "client_id": "resource_owner_client",
+            "client_secret": "014DF517-39D1-4453-B7B3-9930C563627C"
+          },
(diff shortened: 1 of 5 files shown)


config.default.json
```diff
@@ -25,10 +25,30 @@
...
+          "method": "post",
+          "headers": {
+            "Content-Type": "application/x-www-form-urlencoded"
+          }
+        }
       ],
       "default": {
       },
       "required": {
+        "CHECKMARX": "checkmarx.customer.com",
```
assafenso, line 48: @aarlaud actually i dont think we can required any field since the user could add any chosen integration from the list - should we add it to default?

config.default.json
```diff
@@ -25,10 +25,30 @@
...
+            "username": "${CHECKMARX_USERNAME}",
+            "password": "{CHECKMARX_PASSWORD}",
+            "grant_type": "password",
+            "scope": "sast_rest_api",
+            "client_id": "resource_owner_client",
+            "client_secret": "014DF517-39D1-4453-B7B3-9930C563627C"
+          },
+          "method": "post",
+          "headers": {
+            "Content-Type": "application/x-www-form-urlencoded"
```
assafenso, line 41: i think we will need to modify `prepareRequestFromFilterResult` to support conversion from json to `URLSearchParams` to support `x-www-form-urlencoded`

config.default.json
aarlaud: the field required here means that we won't care about the value (we might use it as placeholder/example), but will indicate to the broker that the connection details must have this field define.
It's not completely baked in yet, and based on your use case we're going to have to have a conditional logic to include things whenever applicable, ignore otherwise. We'll combine the 2 I think to achieve the desired behavior.

config.default.json
aarlaud: I am not sure I understand this `support conversion from json to URLSearchParams`. What JSON are you converting into query strings?

assafenso: in order to get checkmarx token we need to send POST request using `application/x-www-form-urlencoded` and inject the user and pass using the `BROKER_VAR_SUB`.
currently the function that replace the Variable substitution sends only JSON body and for this request i need to modify the body as `URLSearchParams`

PR #728, opened 2024-03-01
feat: add ws or http response selection logic

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds the ability to instruct the client to send the response over the websocket connection instead of the HTTPS POST based on header. Gives further control to refine desired behavior.

`x-broker-ws-response` presence alone suffices to indicate the desire to get the response via WS instead of the default HTTPS post.


pavel-snyk (approved):
LGTM.

PR #742, opened 2024-04-11
feat: add raw auth scheme

TheDaemonBarber:
add raw auth scheme option

- [ ] Ready for review

#### What does this PR do?
adding the ability to specify "raw" as the auth scheme allowing the "token" to be used in the header without modification

accept.json example:
{
      "method": "POST",
      "path": "/sbq-my-queue-test/messages",
      "origin": "${SBN_BASE_QUEUE_URL}",
      "auth": {
        "scheme": "raw",
        "token": "SharedAccessSignature ${SBN_MY_QUEUE_TOKEN}"
      }

this is a simple change to the allowed auth themes

Can be used to post messages to an endpoint that uses non-standard auth schemes

We are using the broker to allow access to our private endpoints from the Roadie Scaffolder.  This would allow us to post messages to a private azure service bus instance from the Roadie Scaffolder


aarlaud:
Hey @TheDaemonBarber , it should be fine. Can I ask you to add a couple of test cases in test/functional/server-client.test.ts and test/functional/server-client-universal.test.ts? You'll probably need to add a line in the fixture client accept json and a new uri pattern in the test/setup/test-web-server.ts file to simulate this.

thanks

TheDaemonBarber:
Hi, I've added a unit test, can you confirm if this is sufficient?

aarlaud (approved):

PR #761, opened 2024-05-09
fix: add missing node release key for base image build

aarlaud:
- [x] Ready for review

#### What does this PR do?
Fixes the base image building passing the signature validation step.


dockerfiles/base/Dockerfile.ubi
```diff
@@ -54,9 +54,6 @@ RUN <<EOF
     rpm --erase --nodeps dnf-plugin-subscription-manager
     rpm --erase --nodeps gdb-gdbserver
     rpm --erase --nodeps glibc-common
-    rpm --erase --nodeps glibc-gconv-extra
-    rpm --erase --nodeps glibc-langpack-en
-    rpm --erase --nodeps glibc-minimal-langpack
```
pavel-snyk, line 59: @aarlaud, why do we don't remove it? these deps contain high vulns iirc.

aarlaud: It fails on these saying they are not existent. Maybe the base image changed and they were removed?

pavel-snyk (approved):

dockerfiles/base/Dockerfile.ubi
pavel-snyk: ok, I'll investigate it later. ✅  for this one.

PR #764, opened 2024-05-10
fix: secondary to primary calls to right endpoint

aarlaud:
- [x] Ready for review

#### What does this PR do?
Makes sure to target the right endpoint for request being forwarded from secondary to primary for older client with only one tunnel.

### DIFF
--- a/lib/server/routesHandlers/httpRequestHandler.ts
+++ b/lib/server/routesHandlers/httpRequestHandler.ts
@@ -28,6 +28,7 @@ export const overloadHttpRequestWithConnectionDetailsMiddleware = async (
       localHostname.match(regex)
     ) {
       const url = new URL(`http://${req.hostname}${req.url}`);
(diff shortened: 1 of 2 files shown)


lib/server/routesHandlers/httpRequestHandler.ts
```diff
@@ -28,6 +28,7 @@ export const overloadHttpRequestWithConnectionDetailsMiddleware = async (
       localHostname.match(regex)
     ) {
       const url = new URL(`http://${req.hostname}${req.url}`);
+      url.hostname = req.hostname.replace(/-[0-9]{1,2}/, '');
```
pavel-snyk, line 31: @aarlaud, just to be sure I'm getting the replacement logic: if the hostname is `my-server-123-name` => it will be `my-server-223-name`?

pavel-snyk (approved):

PR #773, opened 2024-05-24
fix: dynamic gids for openshift for rhel-ubi

soniqua:
### What

- Sets directories `/home/snyk` and contents within to be owned by the `snyk` user and the `root` group

### DIFF
--- a/dockerfiles/base/Dockerfile.ubi
+++ b/dockerfiles/base/Dockerfile.ubi
@@ -11,8 +11,6 @@ RUN mkdir -p /tmp/node && \
     curl -fsSLO --compressed "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.xz" && \
     tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /tmp/node --strip-components=1
 
-
-
 FROM node-base as broker-builder
 
 ARG BROKER_VERSION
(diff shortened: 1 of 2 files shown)


dockerfiles/base/Dockerfile.ubi
```diff
@@ -91,8 +91,9 @@ EOF
 
 COPY --from=broker-builder /opt/app-root/src/.npm-global /opt/app-root/src/.npm-global
 COPY --from=node-base /tmp/node/bin/node /usr/bin/node
-COPY --chown=snyk:snyk config.default.json /home/snyk/config.default.json
-COPY --chown=snyk:snyk defaultFilters /home/snyk/defaultFilters
+COPY --chown=snyk:root config.default.json /home/snyk/config.default.json
+COPY --chown=snyk:root defaultFilters /home/snyk/defaultFilters
+RUN chown -R snyk:root /home/snyk
```
pavel-snyk, line 96: Should we add this line here as well (as multiline to reduce amount of Docker layers :))?

https://github.com/snyk/broker/blob/master/dockerfiles/Dockerfile.ubi#L11

soniqua: I think the SCM-specific config is best left in the `dockerfiles/Dockerfile.ubi` file personally - means we don't have to rebuild `/dockerfiles/base/Dockerfile.ubi` container many times (unless I'm misunderstanding the comment!).

pavel-snyk: You're right, we don't rebuild base container many times.

What I mean was: we'll need the same `chown snyk:root` command in `dockerfiles/Dockerfile.ubi` file because we generate accept.json there.

pavel-snyk (approved):

PR #776, opened 2024-06-03
fix: do not install weak deps

soniqua:
### What

- Move from `yum` to `dnf`
- Ignore weak dependencies when building base image

### Why

- Reduce additional packages, limiting surface for vulnerabilities


pavel-snyk (approved):
LGTM.

PR #787, opened 2024-06-26
fix: mitigate vuln SNYK-JS-WS-7266574

aarlaud:
- [x] Ready for review

#### What does this PR do?
Mitigates SNYK-JS-WS-7266574.

Bumping ws requires bumping all primus deps, which can't be done at this time (primus heartbeat mechanism reverses after current version, requiring tight alignement broker client <-> broker server).
Migration will be carried out at a later time once fully on universal broker stack control and version based routing will be introduced.

### DIFF
--- a/lib/common/http/webserver.ts
+++ b/lib/common/http/webserver.ts
@@ -80,6 +80,7 @@ export const webserver = (config, altPort: number) => {
         },
         app,
       );
+  server.maxHeadersCount = 0; // fix https://security.snyk.io/vuln/SNYK-JS-WS-7266574
   server.requestTimeout = process.env.BROKER_WEBSERVER_REQUEST_TIMEOUT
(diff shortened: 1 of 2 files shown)


lib/common/http/webserver.ts
```diff
@@ -80,6 +80,7 @@ export const webserver = (config, altPort: number) => {
         },
         app,
       );
+  server.maxHeadersCount = 0; // fix https://security.snyk.io/vuln/SNYK-JS-WS-7266574
```
soniqua, line 83: q: do we have an upgrade path?

> Upgrade ws to version 5.2.4, 6.2.3, 7.5.10, 8.17.1 or higher.

soniqua (approved):

lib/common/http/webserver.ts
aarlaud: we do but it requires primus and other libraries update which bring a lot of breaking changes, which we can't accomodate at this time.

PR #800, opened 2024-07-25
feat: add url placeholder var swap for ghsa

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds opt-in logic at connection level to enable variable interpolation in urls.
For the cases where a url must contain a value that is only available in the broker client, it is now possible to inject this value simply by having a placeholder value in the filters uris, in addition to the header and body value injections, long time part of the broker client.
This additional capability is required for upcoming integration(s) and will not apply to any current connection without explicit configuration value to opt into this behavior. Cases where needed will be apparent in the filters.

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -287,6 +287,7 @@
         }
       ],
       "default": {
+        "enableUrlVarsSubstitution": true,
         "GITHUB_API": "$GITHUB/api/v3",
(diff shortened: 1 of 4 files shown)


aarlaud:
Closing in favor of a plugin specific approach.

PR #804, opened 2024-07-29
feat: add url placeholder var swap for ghsa in plugin

aarlaud:
- [ ] Ready for review

#### What does this PR do?
Adds opt-in logic in github server app plugin at connection level to enable variable interpolation in urls.
For the cases where a url must contain a value that is only available in the broker client, it is now possible to inject this value simply by having a placeholder value in the filters uris, in addition to the header and body value injections, long time part of the broker client.
Keeping it in the plugin isolates this logic to only relevant type(s).

This approach is favored over the global logic in https://github.com/snyk/broker/pull/800.

### DIFF
--- a/test/unit/plugins/brokerPlugins/github-server-app.test.ts
+++ b/test/unit/plugins/brokerPlugins/github-server-app.test.ts
@@ -3,6 +3,7 @@ import { findProjectRoot } from '../../../../lib/common/config/config';
 import nock from 'nock';
 import { delay } from '../../../helpers/utils';
 import { getConfig } from '../../../../lib/common/config/config';
+import { PostFilterPreparedRequest } from '../../../../lib/common/relay/prepareRequest';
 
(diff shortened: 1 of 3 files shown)


test/unit/plugins/brokerPlugins/github-server-app.test.ts
```diff
@@ -175,4 +176,24 @@ describe('Github Server App Plugin', () => {
...
+      headers: {},
+      method: '',
+    };
+
+    const preequest = await plugin.preRequest(
+      { GITHUB_INSTALLATION_ID: '123' },
+      postFilterPreparedRequest,
+    );
+    expect(preequest).toStrictEqual({
+      url: 'https://hostname/path/test/with/isntallationid/123/rest/of/path',
```
miri-yehezkel, line 194: typo `isntallationid` lines 184 and 194

aarlaud:
closing in favor of #812

PR #822, opened 2024-08-23
Fix/timeout for preflight check

saumilmac:
#### What does this PR do?
This PR adds the ability for request to timeout during the preflight check for better visibility and logging.

### DIFF
--- a/lib/common/http/request.ts
+++ b/lib/common/http/request.ts
@@ -248,11 +248,12 @@ export const makeSingleRawRequestToDownstream = async (
     });
   }
   const httpClient = localRequest.url.startsWith('https') ? https : http;
+  const timeoutMs = req.timeoutMs ?? 0;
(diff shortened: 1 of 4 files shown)


lib/common/http/request.ts
```diff
@@ -287,6 +287,11 @@ export const makeSingleRawRequestToDownstream = async (
         logger.error({ error }, 'Error making request to downstream.');
         reject(error);
       });
+      request.on('timeout', () => {
```
shuaibmunshi, line 290: Out of curiosity, what was causing this loop to fail to timeout when we were trying to debug this with the debugger?

shuaibmunshi (approved):

lib/common/http/request.ts
saumilmac: it was something about `setTimeout` that we were using that didn't play it right

PR #834, opened 2024-09-04
fix: allow one or more matching connection types for plugin enablement

soniqua:
### What

- Adjusts the enablement mechanism for plugins to load if one or more of the defined `applicableBrokerTypes` are present in config

### Why

- Previously _all_ matching broker types must be present

### DIFF
--- a/test/unit/plugins/pluginManager.test.ts
+++ b/test/unit/plugins/pluginManager.test.ts
@@ -34,7 +34,9 @@ describe('Plugin Manager', () => {
     };
     const plugins = await loadPlugins(pluginsFolderPath, clientOpts);
     expect(plugins.get('dummy-multi-1').length).toBeGreaterThanOrEqual(1);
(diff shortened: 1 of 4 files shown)


shuaibmunshi (dismissed):

aarlaud:
I think you're missing a dummy plugin file somewhere for the test

test/unit/plugins/pluginManager.test.ts
```diff
@@ -24,6 +24,19 @@ describe('Plugin Manager', () => {
     expect(plugins.get('dummy')[0].pluginName).toEqual('Dummy Plugin');
   });
 
+  it('should load plugins if at least one supported broker type is present', async () => {
+    const clientOpts = {
+      config: {
+        universalBrokerEnabled: true,
+        supportedBrokerTypes: ['dummy-multi-1'],
+        connections: { 'my connection': { type: 'dummy-multi-1' } },
```
aarlaud, line 32: don't think there is a fixture plugin of that type

PR #862, opened 2024-10-25
chore: add broker token preflight check

soniqua:
- **chore: disable x-forwarded-for headers for gitlab by default**
- **chore: update prettier for**


lib/client/checks/config/index.ts
```diff
@@ -86,3 +87,17 @@ const brokerClientVersionCheck = (config: Config): Check => {
...
   } satisfies Check;
 };
+
+const brokerTokenCheck = (config: Config): Check => {
+  return {
+    id: 'broker-token-validation',
+    name: 'Broker Token Validation Check',
+    enabled: !config.universalBrokerEnabled,
+    check: async function (): Promise<CheckResult> {
+      return await validateBrokerClientVersionAgainstServer(
```
pavel-snyk, line 97:
```suggestion
      return await validateBrokerToken(
```

I guess it should be the different method here.

soniqua: Good catch. Not sure what happened there.

lib/client/checks/utils.ts
```diff
@@ -7,3 +7,10 @@ export const splitStringIntoLines = (
   const lines = inputString.match(regex);
   return lines.join(`\n${linePrefix}`);
 };
+
+export const isUuid = (testValue) => {
+  const regex = new RegExp(
+    `^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`,
+  );
+  return testValue.match(regex);
```
pavel-snyk, line 15: if we bump `uuid` package to 8.3.0 version, we can use `validate()` method: https://github.com/uuidjs/uuid/blob/v8.3.0/src/validate.js#L3

pavel-snyk (approved):
nice, lgtm.

PR #868, opened 2024-11-05
fix: add apprisk rules for classic broker template

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds missing template for apprisk image building in classic broker.

### DIFF
--- a/client-templates/apprisk/.env.sample
+++ b/client-templates/apprisk/.env.sample
@@ -1,0 +1,21 @@
+BROKER_DOWNSTREAM_TYPE_APPRISK=true
+
+# your unique broker identifier
+BROKER_TOKEN=<broker-token>
+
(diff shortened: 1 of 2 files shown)


client-templates/apprisk/.env.sample
```diff
@@ -0,0 +1,21 @@
+BROKER_DOWNSTREAM_TYPE_APPRISK=true
+
+# your unique broker identifier
+BROKER_TOKEN=<broker-token>
+
+# The URL to your artifactory
```
soniqua, line 6: nit: not artifactory

soniqua (changes requested):
A few occurrences of artifactory which should be replaced

soniqua (approved):

PR #872, opened 2024-11-06
chore(circleci): build base images by CICD workflow

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR adds two steps to CICD workflow: building Docker images (Ubuntu and RHEL based). This will help to catch any issues related to building Docker images earlier and not by releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -461,6 +461,21 @@ workflows:
           requires:
             - Install NPM packages
 
+      - build-and-save-docker-image:
+          name: Build base image (Ubuntu)
+          requires:
+            - Install NPM packages
+          dockerfile: dockerfiles/base/Dockerfile
+          nodejs_cycle: "20"
+          project_name: broker
+
+      - build-and-save-docker-ubi-image:
+          name: Build base image (RHEL)
+          requires:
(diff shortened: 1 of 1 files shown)


aarlaud (approved):

.circleci/config.yml
```diff
@@ -461,6 +461,21 @@ workflows:
...
+          nodejs_cycle: "20"
+          project_name: broker
+
+      - build-and-save-docker-ubi-image:
+          name: Build base image (RHEL)
+          requires:
+            - Install NPM packages
+          dockerfile: dockerfiles/base/Dockerfile.ubi
+          project_name: broker-rhel-ubi
+
```
soniqua, line 478: nit: we should require these to pass in the following `release` job

soniqua (changes requested):

.circleci/config.yml
aarlaud: great point

pavel-snyk: makes sense, good catch.

pavel-snyk: done with dffbc9b

soniqua (approved):

PR #873, opened 2024-11-06
Feat/remote rulesets autosync [HYB-714]

aarlaud:
- [ ] Ready for review

#### What does this PR do?
Adds remote ruleset autosyncing from remote and/or local locations.

Moving loaded filters (filter rules loaded into usable function) to it own separate memory object with daily auto refresh and local api endpoint to force download + reload new filters.
Allows to mutate the filters object without impacting configuration and vice versa.

To note: the locations for the respective filters are still defined in the configuration. While changes in the files pointed by these location will be downloaded and reloaded without downtime, changing the location in the config will require restarting the broker.

Other important note: This change applies only to Broker clients (classic and universal alike). Server filters remain static, loaded only at startup.


pavel-snyk (approved):
LGTM, discussed internally.

PR #875, opened 2024-11-11
feat: add header to identify brokered webhooks [SCM-674]

zolamk:
#### What does this PR do?

We thought we could use `snyk-request-id` header to identify brokered scm webhooks but it seems something else is setting that header (perhaps the api gateway), so this PR adds an explicit header to identify brokered webhooks.


aarlaud:
closing in favor of #876.

PR #882, opened 2024-11-14
feat: enable client version control server side

aarlaud:
- [x] Ready for review

#### What does this PR do?
Sets client versions control checks on the server side.
Below a minimal threshold, the connection gets terminated.
Below a minimal soft threshold, a notification is sent to the client indicating client version in use is deprecated and an update is necessary.

Also introduces a notifications mechanisms with clients.


pavel-snyk (approved):
Nice one 🚀

PR #886, opened 2024-12-02
fix: restore ha mode by adding serverId again post refactoring

aarlaud:
- [x] Ready for review

#### What does this PR do?
Fixes a regression breaking the High availability mode for non universal broker.


soniqua (approved):
Confirming this moves the setting of `serverId` out of the universal-broker only config mechanism and ensures it's always set regardless of universal broker mode

PR #888, opened 2024-12-06
fix: add auth headers on all flows [HYB-724]

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds auth headers everywhere needed.

### DIFF
--- a/lib/client/config/configHelpers.ts
+++ b/lib/client/config/configHelpers.ts
@@ -5,6 +5,12 @@ import { getConfig, loadBrokerConfig } from '../../common/config/config';
 import version from '../../common/utils/version';
 import { CONFIGURATION } from '../../common/types/options';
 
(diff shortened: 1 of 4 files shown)


lib/client/config/configHelpers.ts
```diff
@@ -5,6 +5,12 @@ import { getConfig, loadBrokerConfig } from '../../common/config/config';
 import version from '../../common/utils/version';
 import { CONFIGURATION } from '../../common/types/options';
 
+let globalClientOpts: Record<string, any> = {};
```
pavel-snyk, line 8: nitpick: we could use unknown instead of any.

pavel-snyk (approved):

lib/client/config/configHelpers.ts
aarlaud: will fix in the next changes.

PR #912, opened 2025-02-06
fix: add url pattern to the apprisk gitlab whitelist

eduard-panfile-snyk:
- [ ] Ready for review

#### What does this PR do?

This PR is to solve errors that gitlab clients are getting.  Example of these clients are [here](https://snyksec.atlassian.net/browse/IA-872). The issue is related to the fact that a step in sanity checks in apprisk are failing with the following error:

`{
  "message": "blocked",
  "reason": "[Websocket Flow][Blocked Request] Does not match any accept rule",
  "url": "/api/v4/projects/20263346/users"
}`

### DIFF
--- a/client-templates/gitlab/accept.json.sample
+++ b/client-templates/gitlab/accept.json.sample
@@ -49,7 +49,6 @@
       "path": "/api/v4/projects/:project/repository/tree",
       "origin": "https://${GITLAB}"
     },
-
     {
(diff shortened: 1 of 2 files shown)


aarlaud:
Please remove these rules from these files and instead put it in the defaultFilters/apprisk/gitlab.json

PR #925, opened 2025-02-16
Chore/refactor filter and replace

aarlaud:
- [x] Ready for review

#### What does this PR do?

Refactors and separate concerns in code for better modularity.

The main goal here is to separate the filtering function from the variable interpolation function.
Filtering returns either false if no match, or the match rule entry.

Interpolation function will interpolate the config values into the right origin and auth for the request. Any body and/or header var sub replacement is now also part of this interpolation function.

Prepare request is mainly about formatting the request, cleaning up any undesirable headers, etc.

Request preparation is integral part of the broker workload domain. The filtering + variable interpolation are not specific to the broker use case and will be used more generically.

The refactoring is a noop functionality wise. All classic tests and QA remain unchanged.


lib/common/config/config.ts
```diff
@@ -11,6 +11,16 @@ export const getConfig = () => {
...
 
+export const setConfigKey = (key: string, value: unknown) => {
+  config[key] = value;
+};
+
+export const setConfigAndReturnOriginalConfigForTestOnly = (configObject) => {
+  const originalConfig = Object.assign({}, config);
+  config = configObject;
+  return originalConfig;
+};
```
pavel-snyk, line 22: 🤔  we might move this into test `*.test.ts` files, because it's used only in tests.

aarlaud: done

pavel-snyk (approved):

pavel-snyk (approved):

PR #934, opened 2025-02-23
chore: add graceful client shutdown with signaling to server [HYB-866]

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds graceful client shutdown with advanced shutdown signaling to the server which can then signal upcoming client connection shutdown to the load balancing system.

It's a no-op for normal broker behavior. The rest requires the entire stack.

### DIFF
--- a/lib/hybrid-sdk/common/utils/signals.ts
+++ b/lib/hybrid-sdk/common/utils/signals.ts
@@ -1,19 +1,42 @@
-const timers: NodeJS.Timeout[] = [];
+import { getWebsocketConnections } from '../../client';
+import { log as logger } from '../../../logs/logger';
 
+const timers: NodeJS.Timeout[] = [];
+const GRACE_PERIOD_MS = 12000;
 export const handleTerminationSignal = (callback: () => void) => {
   process.on('SIGINT', () => {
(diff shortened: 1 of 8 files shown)


lib/hybrid-sdk/common/utils/signals.ts
```diff
@@ -1,19 +1,42 @@
-const timers: NodeJS.Timeout[] = [];
+import { getWebsocketConnections } from '../../client';
+import { log as logger } from '../../../logs/logger';
 
+const timers: NodeJS.Timeout[] = [];
+const GRACE_PERIOD_MS = 12000;
```
soniqua, line 5: ref: we'll need to update Helm charts if we don't already support graceful shutdown

soniqua (approved):

lib/hybrid-sdk/common/utils/signals.ts
aarlaud: isn't it the standard k8s procedure of giving 30s by default? or you're saying we should have customizable values here?

soniqua: No, I don't think we need to configure this - more just to check the helm charts don't kill the container too early

PR #941, opened 2025-03-04
feat: unibroker add dynamic param injection on req ctx [HYB-915]

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds support for dynamic parameters injection based on specific context contained in the request.

This addition gives the ability to have request specific parameter changes based on context while using an existing connection. It allows to specify a broker connection configuration once as common base for many integrations, while specific context can inject different values when necessary, based on information passed via header from the server.

This mechanism allows the client to read a context ID to inject/override specific parameters from based on a context ID value passed from the server via header.


pavel-snyk (approved):
LGTM

PR #951, opened 2025-03-14
Chore/add handler for cfg and rules syncing actions

aarlaud:
- [ ] Ready for review

#### What does this PR do?
Introduces a service channel to notify customers of state changes instead of local client polling. A connectionless setup still relies on polling initially and if no tunnels are up. If at least one tunnel is up, the polling gets disabled and client will get update from the server via a websocket tunnel service channel to trigger a remote config fetch + reload. Similarly, separate channel is added for filters reload to avoid relying on basic schedules.


pavel-snyk (approved):
LGTM

PR #973, opened 2025-04-15
feat: add context support to plugins manager [HYB-980]

aarlaud:
- [x] Ready for review

#### What does this PR do?
Adds context support in plugins manager.

The contexts are section under a connection in the connections object retrieved from the backend.
The plugin abstract class brings a new startup context method which plugins must implement, giving the opportunity to mutate the configuration just like the regular startup method, but for each context independently.

Using this startupContext method, each context configuration can be mutated to contain whatever updated field/value. The allowed values will likely need tweaking but will be addressed in a separate PR.
Each context object under the given connection will contain the values set by plugins.

The prerequest method, if used/called, will remain largely unchanged, but will get a configuration object containing overriden values by the context (getting the result of `getConfigForIdentifier(connectionIdentifier,clientOpts.config,contextId)` ). It follows the normal pattern for context injection introduced in #941.

Contexts should be seen as sub objects under connections that are used dynamically for requests with a context id associated, overriding default connection config values by different values stored in that context.
Plugins like github app plugins mutate the config object to store in it temporary credentials retrieved from the gh apps. This PR sets the foundations for plugins to do the same thing by calling a startupcontext method to allow a plugin to do the same for each individual context and store the overriding values under the given context.

image

### DIFF
--- a/lib/broker-workload/websocketRequests.ts
+++ b/lib/broker-workload/websocketRequests.ts
@@ -144,10 +144,12 @@ export class BrokerWorkload extends Workload<WorkloadType.remoteServer> {
         this.websocketConnectionHandler?.socketType,
       );
       if (this.options.config.universalBrokerEnabled) {
+        const contextId = payload.headers['x-snyk-broker-context-id'] ?? null;
         preparedRequest.req = await runPreRequestPlugins(
           this.options,
           this.connectionIdentifier,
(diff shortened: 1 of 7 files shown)


pavel-snyk (approved):
LGTM, I like the tests 🥇.

PR #984, opened 2025-05-08
fix: Adds the CRA_COMPATIBLE_TYPES list to default config [HYB-1007]

shuaibmunshi:
- [ ] Ready for review

#### What does this PR do?
Adds a list of integrations that are compatible with the CRA.

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -45,8 +45,12 @@
     "workload"
   ],
   "CRA_COMPATIBLE_TYPES": [
+    "acr",
     "artifactory-cr",
     "digitalocean-cr",
+    "docker-hub",
+    "ecr",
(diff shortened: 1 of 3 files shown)


config.default.json
```diff
@@ -44,6 +44,16 @@
     "google-artifact-cr",
     "workload"
   ],
+  "CRA_COMPATIBLE_TYPES": [
+    "artifactory-cr",
```
aarlaud, line 48: You're missing some types. Docker hub, ever acr gcr

pavel-snyk: nice catch,
fixed with d056b587d1114e489422ba18baf821d1dfb28ef5

PR #1007, opened 2025-06-27
chore: broker should error correctly if run without default.config.json

shuaibmunshi:
- [x] Ready for review

#### What does this PR do?
Returns an error if the broker is run without `config.default.json` to prevent confusion when the broker is misconfigured at run time.

### DIFF
--- a/lib/hybrid-sdk/common/config/config.ts
+++ b/lib/hybrid-sdk/common/config/config.ts
@@ -18,7 +18,8 @@ export const setConfigKey = (key: string, value: unknown) => {
   config[key] = value;
 };
 
-export const findProjectRoot = (startDir: string): string | null => {
+
+export const findProjectRoot = (startDir: string): string => {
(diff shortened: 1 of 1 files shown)


lib/hybrid-sdk/common/config/config.ts
```diff
@@ -31,7 +31,12 @@ export const findProjectRoot = (startDir: string): string | null => {
     currentDir = path.dirname(currentDir);
   }
 
-  return null;
+  const errorMessage =
+    'Error: config.default.json is missing, please ensure the file exists when running the broker.';
+  logger.error(errorMessage);
```
gclapperton, line 36: If we;re not handling the error, generally we shouldn;t log it, otherwise we;ll get an error logged at every layer of the stack until it is handled

shuaibmunshi: Fair enough, I removed the extra log message

PR #1088, opened 2026-03-03
chore: [ACC-3193] always on HA mode for universal broker

gclapperton:
## What changed

`getServerId()` — which registers this client connection with the Snyk dispatcher and receives back a `server_id` used to route websocket traffic in multi-pod deployments — is now called unconditionally when a universal broker client establishes a websocket connection. Previously it was gated behind `BROKER_HA_MODE_ENABLED == 'true'`. The flag and both conditional blocks in `socket.ts` are removed.

`BROKER_HA_MODE_ENABLED` is also removed from all `config.universaltest*.json` files and from the relevant unit test snapshots in `config.test.ts`.

Legacy clients are not affected — `createWebSocketConnectionPairs` is only called when `universalBrokerEnabled` is true.

## Why

Universal broker works in both single-pod and multi-pod deployments, but we want to make scaling to multiple pods as seamless as possible. Since single-pod deployments can participate in server-side routing (HA mode) without any downside, we default to it — so that customers going from one pod to many only need to increase their replica count, with no config changes required.

## Test infrastructure

With `getServerId()` now unconditional, every test that starts a universal broker client makes a dispatcher request against `API_BASE_URL`. Previously tests were pointing that variable at the broker server or the `TestWebServer`, neither of which speaks the dispatcher API, causing those calls to fail or hit unexpected routes.

This PR introduces `TestApiServer` (`test/setup/test-api-server.ts`): a lightweight Express mock of `api.snyk.io` that handles the two dispatcher endpoints (`/hidden/broker/…` and `/hidden/brokers/…`) and the webhook fallback endpoints that the broker client posts to when the websocket is down. All universal broker functional tests now start a `TestApiServer` alongside the broker server and web server. The webhook routes that were previously on `TestWebServer` are removed from there and live only on `TestApiServer`.


Todai88 (approved):
LGTM. 

Can I just ask how we're expecting to release this? Can we let it soak in our test environment?

lib/hybrid-sdk/client/socket.ts
```diff
@@ -73,18 +73,14 @@ export const createWebSocketConnectionPairs = async (
...
+    socketIdentifyingMetadata.identifier,
+    socketIdentifyingMetadata.clientId,
+  );
+
   if (serverId === null) {
-    if (clientOpts.config.BROKER_HA_MODE_ENABLED == 'true') {
-      logger.warn({}, 'Could not receive server id from Broker Dispatcher.');
-    }
+    logger.warn({}, 'Could not receive server id from Broker Dispatcher.');
     serverId = '';
```
mikedevelops, line 84: How does this affect the routing of HA connections in the event we cannot obtain a server ID from the dispatcher? Should we fail early in this case given we are unable to implement HA mode? It looks as though we create the websocket connections regardless of whether we were able to obtain a valid server ID

gclapperton: This doesn't affect the routing of HA connections, it just turns previously non HA brokers into HA brokers (for UB).

gclapperton: I'm hesitant to change the existing behaviour past that as part of this PR

gclapperton:
> LGTM.
> 
> Can I just ask how we're expecting to release this? Can we let it soak in our test environment?

Yes, I'm definitely in favour of letting this soak. Will hold off merging while we sort out [the new release process](https://snyksec.atlassian.net/wiki/spaces/ACC/pages/4700962830/Broker+Client+Upgrade+and+Release+Cadance)

PR #1104, opened 2026-04-01
feat: retry TCP connection to Broker Server

skyth3r:
- [x] Ready for review

### What does this PR do?

Adds retry logic to the Broker Server POST connection for network errors. Retries up to 3 times before failing for network errors such as `ECONNRESET`, `ENOTFOUND`, `ESOCKETTIMEOUT`, `ECONNREFUSED` and 'ETIMEDOUT'.

These changes should improve the resilience of downstream POST streaming to the Broker Server, especially in unreliable network conditions.

Also updates tests to verify and account for this new retry behaviour

#### Changes made
- Added a set of retryable error codes (`ECONNRESET`, `ENOTFOUND`, `ESOCKETTIMEOUT`, `ECONNREFUSED`, `ETIMEDOUT`) 
- Implemented retry logic (up to 3 retries) with backoff (500ms between attempts)
- Adds error handling to sendData (similar to forwardRequest)
- Updated tests to reduce duplication and cover the retry scenarios


lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -466,7 +494,49 @@ class BrokerServerPostResponseHandler {
...
+              errorCode: (this.#retryableError as any)?.code,
+            },
+            'Retrying POST connection to Broker Server',
+          );
+          await delay(POST_RETRY_DELAY_MS);
+          this.#buffer = this.#createBuffer();
+          this.#retryableError = undefined;
+        }
+
+        await this.#initHttpClientRequest();
```
gclapperton: Correct me if I'm wrong, but does init request not just create a request object and not actually make (or attempt to make) a connection to the broker server?

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -171,197 +189,256 @@ class BrokerServerPostResponseHandler {
...
-                readableHighWaterMark: this.#buffer.readableHighWaterMark,
-                writableLength: this.#buffer.writableLength,
-                writableHighWaterMark: this.#buffer.writableHighWaterMark,
-              },
-              stackTrace: new Error('stacktrace generator').stack,
-            },
-            'Upstream request to Broker Server timed out',
-          );
-          this.#brokerSrvPostRequestHandler?.destroy();
-          this.#buffer.end(timeoutError.message);
```
clwiseman, line 297: This was removed because it is redundant and this would cause the retry mechanism to fail. The buffer is already ended in `handleRequestError` or as part of the cleanup with using `pipeline`.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -171,197 +189,256 @@ class BrokerServerPostResponseHandler {
...
-              errDetails: e,
-              stackTrace: new Error('stacktrace generator').stack,
-              streamingID: this.#streamingId,
-              requestId: this.#requestId,
-              brokerServerUrl: this.#config.brokerServerUrl,
-              ...extractNetworkErrorDetails(e),
-            },
-            'received error sending data via POST to Broker Server',
-          );
-          this.#buffer.end(e.message);
```
clwiseman, line 272: This was removed because it is redundant and this would cause the retry mechanism to fail. The buffer is already ended in `handleRequestError` or as part of the cleanup with using `pipeline`.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: This is outdated so it might be more clear now, but It does actually establish the connection! When it calls `client.request`, it starts to establish the connection asynchronously. While that starts up we add a bunch listeners and then we await the Promise until it either listens to a socket or it errors out (and attempts a retry).

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -466,7 +533,9 @@ class BrokerServerPostResponseHandler {
         status,
         headers: response.headers,
       });
+
       await this.#initHttpClientRequest();
```
gclapperton, line 554: is this function async? I know the signature says that it is, but I'm not sure why

gclapperton, line 554: Can we type this?

gclapperton, line 554: If the server is experiencing issues, would it be better to have some sort of backoff? (potentially with jitter)

gclapperton, line 554: Might be a good idea to pull the retry logic out. We have the same logic in two places and these fns are already confusing

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -538,7 +608,16 @@ class BrokerServerPostResponseHandler {
...
-    await this.#initHttpClientRequest();
+    try {
+      await this.#initHttpClientRequest();
+    } catch (err) {
+      this.#logger.error(
+        { errorDetails: err },
+        'Error establishing connection for POST to Broker Server',
+      );
+      this.#buffer.destroy();
+      return;
```
gclapperton, line 636: Might we want to rethink this? Having the pipeline error would make this retry logic more straight forward

gclapperton, line 636: Not sure if this is just js, but feels mucky. Is there a cleaner way to do this?

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
gclapperton: ah, ok, so the request does not get sent until `req.end()` is called, but it does establish a TCP connection with the server 🙏

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: What would we type? This function doesn't return anything.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: How so? The retry is not for the pipeline part of the function - it's for establishing the TCP connection.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: It is async! Within this function we are awaiting the TCP connection to be established: https://github.com/snyk/broker/pull/1104/changes#diff-6364f20b90ff315ea447603632a8827af8c0731a7c8a095f2249a3976f759a76R384-R396

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: The connection isn't established when `pipeline` is called, it's established when the Promise is awaited in the init function: https://github.com/snyk/broker/pull/1104/changes#diff-6364f20b90ff315ea447603632a8827af8c0731a7c8a095f2249a3976f759a76R384-R396

Do you want to sync on this? We can talk it through.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: I'm not sure what you mean. Do you mean the two functions `sendData` and `forwardRequest`? They are used in two different scenarios - one is for regular HTTP requests and one is for streaming requests.

The retry logic is not in two places. It's just in the init function which is a shared function of both.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: Correction: The request is not sent until `pipeline` is called.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: Which part is mucky? Closing the buffer?

saumilmac (approved):

PR #1108, opened 2026-04-14
chore: add new websocket event handling

Todai88:
The purpose of this change is to improve our overall telemetry surface.

The new metric has a human-readable event as a label, rather than the "raw" websocket event. This hopefully makes it clearer for our downstream users what's specifically going on.

Compare "closing" to "server_requested_close".

Also:
- renames the closeHandler to reportWebSocketClosureEvent - this is much closer to what we actually are doing. The code isn't handling any events.
- implements the reportWebSocketClosureEvent function to emit new telemetry
- adds tests to ensure existing and current functionality continues working

### DIFF
--- a/lib/hybrid-sdk/client/socket.ts
+++ b/lib/hybrid-sdk/client/socket.ts
@@ -10,7 +10,7 @@ import {
 import { identifyHandler } from './socketHandlers/identifyHandler';
 import { errorHandler } from './socketHandlers/errorHandler';
 import { openHandler } from './socketHandlers/openHandler';
-import { closeHandler } from './socketHandlers/closeHandler';
+import { reportWebSocketClosureEvent } from './socketHandlers/reportWebSocketClosureEvent';
 import { IdentifyingMetadata, Role, WebSocketConnection } from './types/client';
 import { requestHandler } from './socketHandlers/requestHandler';
 import { chunkHandler } from './socketHandlers/chunkHandler';
@@ -273,6 +273,7 @@ export const createWebSocket = (
   }
 
   websocket.on('incoming::error', (e) => {
+    metricsClient.recordWebsocketLifecycleEvent('connection_error', identifyingMetadata.role);
     websocket.emit('error', { type: e.type, description: e.description });
   });
 
@@ -325,9 +326,9 @@ export const createWebSocket = (
   websocket.on('notification', notificationHandler);
   websocket.on('error', errorHandler);
 
-  websocket.on('open', () =>
-    openHandler(websocket, localClientOps, identifyingMetadata, metricsClient),
-  );
+  websocket.on('open', () => {
+    openHandler(websocket, localClientOps, identifyingMetadata, metricsClient);
+  });
 
   websocket.on('service', serviceHandler);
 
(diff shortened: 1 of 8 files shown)


lib/hybrid-sdk/client/socket.ts
```diff
@@ -325,9 +326,9 @@ export const createWebSocket = (
   websocket.on('notification', notificationHandler);
   websocket.on('error', errorHandler);
 
-  websocket.on('open', () =>
-    openHandler(websocket, localClientOps, identifyingMetadata, metricsClient),
-  );
+  websocket.on('open', () => {
+    openHandler(websocket, localClientOps, identifyingMetadata, metricsClient);
```
mikedevelops, line 330: Is it intentional that the `open` event is not considered a lifecycle event?

lib/hybrid-sdk/client/socket.ts
```diff
@@ -336,7 +337,23 @@ export const createWebSocket = (
   });
 
   websocket.on('close', () => {
-    closeHandler(websocket, localClientOps, identifyingMetadata, metricsClient);
+    reportWebSocketClosureEvent(websocket, localClientOps, identifyingMetadata, metricsClient, 'connection_lost');
```
mikedevelops, line 340: How did we get to these human readable names? Are we introducing a layer of confusion by not sticking with the events the websocket client uses? They already seem pretty human readable to me...

test/unit/client/metrics/metricsClient.test.ts
```diff
@@ -419,6 +420,82 @@ describe('client/metrics', () => {
...
+          (dp) => dp.attributes['event'] === 'connection_timed_out',
+        )?.value,
+      ).toBe(1);
+      expect(
+        dataPoints.find(
+          (dp) => dp.attributes['event'] === 'connection_error',
+        )?.value,
+      ).toBe(1);
+    });
+
```
mikedevelops: nit: this could be tidied up a bit with a helper function for asserting the data points

lib/hybrid-sdk/client/socket.ts
Todai88: We should standardise on lifecycleEvent, but the openHandler already emits other connection-related metrics.

lib/hybrid-sdk/client/socket.ts
Todai88: Unfortunately, some of the events can be slightly misleading. 
Like `closing` -- one could be excused if one believed it was emitted as a "client is intending to close", but it's actually an event emitted by the `server` ahead of actually closing the connection.

test/unit/client/metrics/metricsClient.test.ts
Todai88: Great idea! 

Let me sort that out!

lib/hybrid-sdk/client/socket.ts
mikedevelops: This is a good example IMO:

>  but it's actually an event emitted by the server ahead of actually closing the connection.

Does `connection_lost` make this less ambiguous? Seems like it's making an assumption about why the event fired vs what event fired?

Todai88: Unfortunately, I think the ambiguity there is necessary for _now_. 
There are a lot of reasons why a connection may be lost, and we don't currently have a way of determining if the client itself closed the connection, or if some network jitters cause it. 

It's not technically infeasible to make the client communicate its intent ahead of it firing off a WS close event, which would give us two individual close events "client initiated close" and "server intiated close", and could give a bit more context on why something failed.

But I don't believe we want to baloon this PR to include that change.

mikedevelops (approved):

PR #1127, opened 2026-05-15
feat: guarantee `snyk-request-id` on every inbound HTTP request [ACC-3093]

gclapperton:
[ACC-3093](https://snyksec.atlassian.net/browse/ACC-3093)

## Summary

Today, broker logs frequently show `requestId: ''` because no middleware ever sets a request ID — if the upstream caller omits the header, every log line for that request is untraceable. SCM webhooks, container registry agents, and Kubernetes probes never send `snyk-request-id`, yet broker tries to log it for them.

This PR guarantees every request handled by the broker webserver (both client and server modes) has a stable, valid UUID for its lifetime:

- If the inbound request carries a recognised header with a valid UUID, that value is used (first match wins across `x-akamai-request-id` → `x-request-id` → `snyk-request-id` → `x-github-delivery` → `x-gitlab-event-uuid`)
- Otherwise a fresh UUIDv4 is generated
- The resolved value is echoed back to the caller in the `snyk-request-id` response header

The middleware is mounted before `bodyParser` so it runs on every route — `/healthcheck`, `/webhook/*`, `/api/v*`, etc. It fails open: if UUID generation fails, a warning is logged and the request continues.

All HTTP-path reads of `req.headers['snyk-request-id']` across the codebase are replaced with the typed `req.requestId` accessor. Where downstream WebSocket-path code still reads the raw header, `req.headers['snyk-request-id']` is backfilled from `req.requestId` as a bridge until those reads are updated in a subsequent PR.

## Test plan

- [ ] 17 unit tests covering: each candidate header accepted in isolation, full priority ordering across all 5 headers, UUID validation, nil UUID rejection, fallback generation, custom options, fail-open error path
- [ ] 2 new functional tests: synthesised ID present on response when caller omits it, supplied ID preserved when caller provides one
- [ ] Full unit suite passes with no regressions (610 tests)
- [ ] Build type-checks clean

[ACC-3093]: https://snyksec.atlassian.net/browse/ACC-3093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/hybrid-sdk/common/http/middleware/requestId.ts
+++ b/lib/hybrid-sdk/common/http/middleware/requestId.ts
@@ -1,0 +1,61 @@
+import { randomUUID } from 'crypto';
+import type { Request, Response, NextFunction, RequestHandler } from 'express';
+import { log as logger } from '../../../../logs/logger';
+import { isUUID } from '../../utils/uuid';
+
+/** Inbound headers inspected for an existing request ID, in priority order. */
+const DEFAULT_INHERITED_HEADERS: ReadonlyArray<string> = [
+  'x-akamai-request-id',
+  'x-request-id',
+  'snyk-request-id',
+  'x-github-delivery',
+  'x-gitlab-event-uuid',
+];
+
+/** Response header used to echo the resolved request ID back to the caller. */
+const DEFAULT_RESPONSE_HEADER = 'snyk-request-id';
+
+export interface RequestIdHeaderOptions {
+  /** Headers to inspect for an existing request ID, checked in order. */
+  incomingHeaders?: ReadonlyArray<string>;
+  /** Response header name used to echo the resolved request ID back to the caller. */
+  responseHeader?: string;
+}
+
+/**
+ * Express middleware that guarantees `req.requestId` is set to a valid UUID
+ * string for every request handled by this server.
+ *
+ * Resolution order:
+ * 1. The first header in `incomingHeaders` whose value is a valid, non-nil UUID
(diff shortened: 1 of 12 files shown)


lib/hybrid-sdk/common/http/middleware/requestId.ts
```diff
@@ -0,0 +1,61 @@
+import { randomUUID } from 'crypto';
+import type { Request, Response, NextFunction, RequestHandler } from 'express';
+import { log as logger } from '../../../../logs/logger';
+import { isUUID } from '../../utils/uuid';
+
+/** Inbound headers inspected for an existing request ID, in priority order. */
+const DEFAULT_INHERITED_HEADERS: ReadonlyArray<string> = [
+  'x-akamai-request-id',
+  'x-request-id',
+  'snyk-request-id',
```
mikedevelops, line 10: nit: should this be top priority? This is the paved path header that our WASM plugins (gateway, cerberus) will set

mikedevelops (approved):

lib/hybrid-sdk/common/http/middleware/requestId.ts
gclapperton: This is consistent with [snyk/rest-node-libs](https://github.com/snyk/rest-node-libs/blob/main/src/request-id/middleware.ts#L9)

PR #1132, opened 2026-05-19
fix: retry oauth token refresh

clwiseman:
- [ ] Ready for review

#### What does this PR do?

Currently, if an oauth token fails to refresh, the broker client gets into a bad state that is only discovered when the token is attempted to be used - HTTP requests back to Snyk will continue to fail with a 401.

This implements a fix to trigger a refresh of the oauth token in two scenarios:
1. If auth refresh fails a retry is triggered.
2. If the HTTP request to Snyk fails with a 401, the request will fail but the auth refresh is triggered so a reattempt will succeed. Due to the nature of HTTP post streaming, it's best not to retry the HTTP request itself with a fresh token and risk the data being malformed.

### DIFF
--- a/lib/hybrid-sdk/client/auth/oauth.ts
+++ b/lib/hybrid-sdk/client/auth/oauth.ts
@@ -1,3 +1,4 @@
+import { v4 as uuid } from 'uuid';
 import { PostFilterPreparedRequest } from '../../../broker-workload/prepareRequest';
 import { makeRequestToDownstream } from '../../http/request';
 import { log as logger } from '../../../logs/logger';
@@ -24,6 +25,7 @@ export async function fetchAndUpdateJwt(
   apiHostname: string,
   clientId: string,
   clientSecret: string,
+  requestId: string = uuid(),
 ) {
   try {
     const data = {
@@ -35,7 +37,10 @@ export async function fetchAndUpdateJwt(
 
     const request: PostFilterPreparedRequest = {
       url: `${apiHostname}/oauth2/token`,
-      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+        'Snyk-Request-Id': requestId,
+      },
       method: 'POST',
       body: formData.toString(),
     };
@@ -59,9 +64,20 @@ export async function fetchAndUpdateJwt(
     });
     return { expiresIn: expiresIn, authHeader: `${type} ${jwt}` };
   } catch (err) {
-    logger.error({ err }, 'Unable to retrieve JWT');
+    logger.error({ err, requestId }, 'Unable to retrieve JWT');
   }
 }
+
+const JWT_REFRESH_RETRY_ON_FAILURE_MS = 30_000;
+
+let refreshTimer: ReturnType<typeof setTimeout> | null = null;
+
+export const stopJwtRefresh = () => {
+  if (refreshTimer) {
+    clearTimeout(refreshTimer);
+    refreshTimer = null;
+  }
+};
 
 const refreshJwt = async (
   clientConfig,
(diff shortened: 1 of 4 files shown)


clwiseman:
> [**Negative Timeout Potential**](https://github.com/snyk/broker/pull/1132/changes#diff-eccfad76a826c97a955d2c481d89a962013299deb8ed562df6be6d037d99ef77R128-R129) 🟡 [minor]
> In `refreshJwt`, the calculation `(newJwt.expiresIn - 60) * 1000` for `nextDelayMs` can result in a negative value if the OAuth provider returns a token with a very short TTL (less than 60 seconds). While `setTimeout` treats negative values as 1ms, this could lead to an tight infinite refresh loop if the token provider consistently issues short-lived tokens or if there is clock skew.

If it does have a single instance of a tight refresh, it won't be a loop because the next will be the default of 60m TTL.

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
```diff
@@ -370,6 +374,16 @@ class BrokerServerPostResponseHandler {
...
           });
+          if (r.statusCode === 401 && this.#config.universalBrokerGa) {
+            logger.warn(
+              {
+                requestId: this.#requestId,
+                streamingId: this.#streamingId,
+              },
+              'Broker server rejected POST with 401; triggering OAuth token refresh',
+            );
+            oauthEvents.emit(OAUTH_TOKEN_REJECTED_EVENT);
```
mikedevelops, line 385: What's the rationale of using an event emitter here? Why not just call a refresh function from the oauth package?

lib/hybrid-sdk/client/auth/oauth.ts
```diff
@@ -59,22 +68,55 @@ export async function fetchAndUpdateJwt(
     });
     return { expiresIn: expiresIn, authHeader: `${type} ${jwt}` };
   } catch (err) {
-    logger.error({ err }, 'Unable to retrieve JWT');
+    logger.error({ err, requestId }, 'Unable to retrieve JWT');
   }
 }
 
+const JWT_REFRESH_RETRY_ON_FAILURE_MS = 30_000;
```
mikedevelops, line 75: 30s to retry a failed request to exchange seems a bit high? Should we retry with some backoff? Presumably broker is in a terminal state at this point?

lib/hybrid-sdk/client/auth/oauth.ts
```diff
@@ -98,15 +144,22 @@ export const setfetchAndUpdateJwt = async (
...
+    requestId,
   );
-  logger.debug({}, 'Setting auth updater.');
+  logger.debug({ requestId }, 'Setting auth updater.');
   if (newJwt) {
-    setTimeout(async () => {
-      refreshJwt(clientConfig, clientId, clientSecret, metricsClient);
+    refreshTimer = setTimeout(() => {
+      startRefreshJwt(clientConfig, clientId, clientSecret, metricsClient);
     }, clientConfig.AUTH_EXPIRATION_OVERRIDE ?? (newJwt.expiresIn - 60) * 1000);
```
mikedevelops, line 163: This is interesting. I would have thought we'd handle refresh at the point the server tells us the token is invalid – I guess we are trying to pre-emptively refresh here?

lib/hybrid-sdk/http/downstream-post-stream-to-server.ts
clwiseman: We don't have access to the client ID and secret from this part of the code to call the auth refresh function from here - it's loaded from config when the client is started and the auth refreshes on a timer.

I don't love the event as a solution but I thought this was preferable to loading the client ID and secret configuration into the HTTP streaming handler since it should not need to know about those sensitive values.

lib/hybrid-sdk/client/auth/oauth.ts
clwiseman: This has me wanting to think more about how we are using the refresh token 🤔 I've been trying to work around the existing set-up but possibly a bigger refactor is warranted here.
We currently do a pre-emptive refresh every 55 minutes, and the token expires after 60 minutes. I'm trying to harden this a bit so that we attempt to refresh if the token is invalid. Typically in an OAuth2 library it would behave as you've said - the library provides an HTTP client that auto-refreshes the token if the token fails with 401. This would be a more robust approach that might end up simplifying the code. Let me think on it 🤔

clwiseman:
Closed in favour of https://github.com/snyk/broker/pull/1141

lib/hybrid-sdk/client/auth/oauth.ts
clwiseman: Updated! This is such a huge change I just opened a new PR https://github.com/snyk/broker/pull/1141

PR #1137, opened 2026-05-20
refactor(logs): rebalance network and forward-compat log levels (TDD)

Todai88:
PR description

  ## Why

  Four log sites in the broker client were set to levels that would've/should've asked oncall to act when no action was warranted, or that flagged a benign outcome as "degraded":

  - `TransportError` on every websocket blip fired at ERROR. On a stable network these recover within milliseconds and the broker keeps working. The unrecoverable case (reconnect attempts exhausted) already escalates to ERROR + `process.exit(1)` in a different file, so the loud channel is preserved for the genuinely-broken case.
  - The `'ECONNRESETs Catch all:'` line lives inside an `uncaughtException` handler that **swallows** the error rather than crashing. That's an explicit admission that the event is expected and recoverable — yet it logged at ERROR. The metrics counter `metricsClient.recordUncaughtException(error.code)` two lines above already captures this for rate-based alerting.
  - `'Unknown service message received.'` fires when the broker server sends a command the client doesn't have a case for. That's the textbook forward-compat scenario (newer server, older client during a rollout). ERROR-per-unknown-command turns a fleet-wide rollout into a phantom incident.
  - `'Server did not advertise received-post-streams capability — falling back to legacy streaming.'` fires once per connection when capability negotiation chooses the legacy path. Both outcomes (advertise / don't advertise) are valid; neither is degraded. WARN was the wrong level pillar; INFO is the right one (lifecycle / state-change).

  Combined with the earlier PRs in this audit (per-request INFO trio demoted; lifecycle events promoted; plugin config redaction), the default-level log now reads as a coherent lifecycle/status narrative and ERROR carries meaningful "page me" weight again.

  ## What

  4 one-line level flips + 1 testability extraction.

  | File | Flip | Message |
  |---|---|---|
  | `lib/hybrid-sdk/client/socketHandlers/errorHandler.ts` | ERROR → WARN | `Failed to connect to broker server.` |
  | `lib/hybrid-sdk/client/index.ts` | ERROR → WARN | `ECONNRESETs Catch all:` |
  | `lib/hybrid-sdk/client/socketHandlers/serviceHandler.ts` | ERROR → WARN | `Unknown service message received.` |
  | `lib/hybrid-sdk/requestsHelper.ts` | WARN → INFO | `Server did not advertise received-post-streams capability - falling back to legacy streaming.` |

  Extraction: `lib/hybrid-sdk/client/index.ts` previously embedded the ECONNRESET-handling body inside an anonymous `process.on('uncaughtException', ...)` callback. Now extracted into a named export `handleUncaughtException(error, metricsClient)` and the `process.on` line calls it. No behaviour change — same code path, just reachable from a unit test. **Caveat**: the unit test verifies the extracted function's level,
  not that it's still wired up as the uncaughtException handler — a code-review concern but not something a unit test can reasonably cover.

### DIFF
--- a/lib/hybrid-sdk/client/index.ts
+++ b/lib/hybrid-sdk/client/index.ts
@@ -81,34 +81,9 @@ export const main = async (clientOpts: ClientOpts) => {
     // NOTE: If main() is called more than once, additional listeners will
     // accumulate after this point — each call registers a new handler on top.
     process.removeAllListeners('uncaughtException');
-    process.on('uncaughtException', (error) => {
-      const isEconnreset = error.message === 'read ECONNRESET';
-      metricsClient.recordUncaughtException(
-        (error as NodeJS.ErrnoException).code ?? error.message ?? 'unknown',
-      );
-      if (isEconnreset) {
-        logger.error(
-          { msg: error.message, stackTrace: error.stack },
-          'ECONNRESETs Catch all:',
-          error.message,
-        );
-      } else {
-        metricsClient.recordProcessExit('uncaught_exception');
-        logger.error(
-          { msg: error.message, stackTrace: error.stack },
-          'Uncaught exception:',
-          error.message,
-        );
-        const flushTimeout = new Promise<void>((resolve) =>
-          setTimeout(resolve, 2000),
-        );
-        Promise.race([metricsClient.forceFlush(), flushTimeout])
-          .catch((err) =>
-            logger.warn({ err }, 'Failed to flush metrics before exit.'),
-          )
-          .finally(() => process.exit(1));
-      }
-    });
+    process.on('uncaughtException', (error) =>
+      handleUncaughtException(error, metricsClient),
+    );
 
     clientOpts.config.API_BASE_URL =
       clientOpts.config.API_BASE_URL ??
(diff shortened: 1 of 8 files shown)


lib/hybrid-sdk/client/index.ts
```diff
@@ -297,3 +272,38 @@ export const main = async (clientOpts: ClientOpts) => {
...
+      error.message,
+    );
+    const flushTimeout = new Promise<void>((resolve) =>
+      setTimeout(resolve, 2000),
+    );
+    Promise.race([metricsClient.forceFlush(), flushTimeout])
+      .catch((err) =>
+        logger.warn({ err }, 'Failed to flush metrics before exit.'),
+      )
+      .finally(() => process.exit(1));
```
mikedevelops, line 307: why are doing this? Seems odd to exit the process from this function? Our healthchecks should take care of that?

Todai88: That's a very good question; all I've done here is extract the code from the main process into a new function so we can actually test the behaviour. 

We could defer the decision to restart the binary based on health checks, but I don't think this PR is the right place for that discussion. Though that's a worthwhile discussion to have. 

E.g: bubbling up the degraded state to k8s and let k8s reschedule, or let the binary exit the process when it gets in a problematic state and re-establish all connections. 

Are you OK with us deferring this to later?

mikedevelops: yeah if this is the existing behaviour then that works for me.

mikedevelops (approved):

PR #1141, opened 2026-05-20
fix: universal broker oauth token refresh

clwiseman:
- [x] Ready for review

#### What does this PR do?

Replaces the custom token-refresh loop (setTimeout-driven, mutable authConfig singleton) with a simple-oauth2 ClientCredentials client. Callers fetch tokens lazily via getAccessToken() / getCachedAccessToken() and the library handles expiry. Previously, there was a bug with the token refresh due to a lack of retries when the token refresh failed.

When the server rejects a request with 401, the cached OAuth token may be stale (e.g. rotated or revoked server-side). Invalidate it and retry the request once with a freshly fetched token before failing.

Also updates the websocket renewal interval default to 10 minutes (still overridable via AUTH_EXPIRATION_OVERRIDE), since the cached token now refreshes on demand rather than on a timer.


clwiseman:
> [**Fixed Renewal Interval**](https://github.com/snyk/broker/pull/1141/changes#diff-5f45be3f7a1a2ce506d53ca700d15910ee0ded32832e7a511680b953fc6459b6R214-R215) 🟠 [major]
> The WebSocket renewal interval is now hardcoded to a default of 10 minutes. Previously, it was dynamically calculated based on the actual token lifetime (`expiresIn - 60`). If the OAuth provider issues tokens with a TTL shorter than 10 minutes, the session will expire at the server before the client attempts a renewal, leading to systematic 401 errors and socket disconnections.

This is not an issue. It's based on the invalid assumption that the token lifetime and the auth refresh interval should be the same, and this is simply not true. Default tokens are 1 hour so this should not happen, but if it does the auth refresh will retry by design if it gets a 401.

 
> [**Missing Auth Header during Refresh**](https://github.com/snyk/broker/pull/1141/changes#diff-5f45be3f7a1a2ce506d53ca700d15910ee0ded32832e7a511680b953fc6459b6R182-R185) 🟠 [major]
> In `buildExtraHeaders`, the `Authorization` header is omitted if `getCachedAccessToken()` returns `undefined`. `getCachedAccessToken` returns `undefined` whenever the token is `expired(0)`, even though it triggers a background refresh. Since `reconnect scheduled` is a synchronous event, the client will immediately attempt a reconnection without any Authorization header, which is guaranteed to fail at the server. This causes transient, avoidable reconnection failures during the period the background refresh is in flight.

This is a feature not a bug. This code path should never be triggered because the OAuth client is initialized first., but if it is we do want the container to restart because the client cannot communicate with the server without that token.

> [**Latency in Streaming Path**](https://github.com/snyk/broker/pull/1141/changes#diff-6364f20b90ff315ea447603632a8827af8c0731a7c8a095f2249a3976f759a76R210-R210) 🟡 [minor]
> The call to `await getAccessToken()` is now part of the serial initialization of every POST stream request to the Broker Server. If the token is within the expiry threshold, this call blocks the start of the stream while it performs a network request to the OAuth provider. For high-throughput streaming environments, this adds significant latency to stream creation that could be avoided by using `getCachedAccessToken` or ensuring background refreshes do not block the hot path.

Since the token is required for the POST streaming request, this function must be blocking.

saumilmac (approved):

PR #1151, opened 2026-05-22
feat: downstream x-request-id mirror + webhook HTTP fallback fix

gclapperton:
## Context

This PR is part of a multi-PR effort to guarantee every request handled by broker carries a stable, traceable request ID end-to-end.

- PR #1127: HTTP inbound — every request hitting broker's Express server gets `snyk-request-id` assigned by middleware and echoed in the response.
- PR #1142: WebSocket inbound — every WS `request` event payload has `snyk-request-id` backfilled before it reaches `BrokerWorkload.handler`.

This PR extends that guarantee outward to the customer system and close a gap in the webhook fallback path.

## What problem does this solve?

**Gap 1 — Downstream tracing.** When broker fires an outbound HTTP request to a customer system (GitHub, GitLab, Jira, etc.), it sends `snyk-request-id` but never the more standard `x-request-id`. Tracing systems and proxies on the customer side that key off `x-request-id` cannot correlate the request back to anything in Snyk's traces.

**Gap 2 — Webhook HTTP fallback.** The normal webhook flow (SCM → broker-client → WebSocket → broker-server → Snyk API) is fully covered by the other two PRs But when the WebSocket is unavailable, the broker-client falls back to a direct HTTP call to the Snyk API. In that fallback path, whatever `snyk-request-id` the Snyk API chose to echo could overwrite the UUID the broker had already assigned — breaking the invariant that the broker's ID is the correlation handle for the full round trip.

### DIFF
--- a/lib/hybrid-sdk/common/http/middleware/requestId.ts
+++ b/lib/hybrid-sdk/common/http/middleware/requestId.ts
@@ -2,6 +2,7 @@ import { randomUUID } from 'crypto';
 import type { Request, Response, NextFunction, RequestHandler } from 'express';
 import { log as logger } from '../../../../logs/logger';
 import { isUUID } from '../../utils/uuid';
+import { extractBrokerTokenFromUrl, maskToken } from '../../utils/token';
 
 /** Inbound headers inspected for an existing request ID, in priority order. */
 const DEFAULT_INHERITED_HEADERS: ReadonlyArray<string> = [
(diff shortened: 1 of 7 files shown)


lib/hybrid-sdk/common/http/middleware/requestId.ts
```diff
@@ -50,8 +51,12 @@ export const setRequestIdHeader = (
         req.requestId = inherited;
       } else {
         req.requestId = randomUUID();
+        const brokerToken = extractBrokerTokenFromUrl(req.url);
+        const safeUrl = brokerToken
+          ? req.url.replaceAll(brokerToken, maskToken(brokerToken))
+          : req.url;
         logger.debug(
-          { method: req.method, url: req.url },
```
gclapperton, line 54: introduced in the last PR

shuaibmunshi (approved):

PR #1155, opened 2026-05-25
refactor: introduce typed requestId accessor across proxy request path

gclapperton:
## Summary

Direct reads of \`snyk-request-id\` from header bags were scattered across 14 files, creating several problems: easy to miss sites when tracing a request, silent empty-string fallbacks on paths where the header wasn't propagated, and no single place to reason about where the ID comes from.

This PR centralises the concern:

- \`RequestPayload\` gains a required \`requestId: string\` field, populated by the WS backfill in \`forwardWebsocketRequest\` alongside the \`snyk-request-id\` header
- \`PostFilterPreparedRequest\` gains an optional \`requestId?: string\` field, populated explicitly from \`payload.requestId\` on the proxy path
- All direct \`headers['snyk-request-id']\` reads in request handlers and filter logic are replaced with the typed accessor
- Both \`websocket.send\` call-sites in \`clientRequestHelpers\` now include \`requestId\` in the payload so \`forwardWebsocketRequest\` receives it directly rather than synthesising a fresh UUID
- The three \`make*RequestToDownstream\` functions resolve \`requestId\` (inherited or synthesised) and write it into the outgoing \`snyk-request-id\` header — so every outbound HTTP call carries the header, including self-initiated broker requests (auth, health checks, config fetches) that previously had no request ID at all
- \`BrokerWorkload\` log context now uses \`payload.requestId\` as the authoritative source rather than re-reading from the header spread, eliminating a potential divergence if headers are mutated downstream
- Logging gaps fixed: \`filtersAsync\` rule-match debug, \`broker-middleware\` broker-type warn (was nested under an extra key), two \`httpRequestHandler\` debug calls, and three \`connectionStatusHandler\` log calls (debug, error, warn) were all missing \`requestId\` despite having it in scope
- Comments added at the three legitimate direct header read sites explaining why the typed accessor cannot be used (Primus authorize callback runs pre-middleware, service-channel WS messages have no Express request, outbound API response read has no typed accessor on the response type)

Three sites that cannot use the accessor are left as direct header reads with explanatory comments.

## Test plan

- [ ] All unit test suites pass (\`npm run test:unit\`)
- [ ] Typecheck passes (\`npx tsc --noEmit\`)
- [ ] New tests cover: \`payload.requestId\` backfill in \`forwardWebsocketRequest\`, \`requestId\` propagation through \`prepareRequest\`, typed accessor usage (including UUID synthesis fallback) in all three \`make*RequestToDownstream\` functions, \`requestId\` field in WS payloads sent from \`clientRequestHelpers\`, and \`payload.requestId\` as authoritative source in \`BrokerWorkload\` log context
- [ ] Post-refactor grep confirms only the expected cannot-refactor read sites remain: \`grep -rn "headers\['snyk-request-id'\]" lib/ --include="*.ts"\`

### DIFF
--- a/lib/broker-workload/prepareRequest.ts
+++ b/lib/broker-workload/prepareRequest.ts
@@ -29,6 +29,7 @@ export interface PostFilterPreparedRequest {
   body?: any;
   timeoutMs?: number;
   tlsOptions?: Record<string, boolean | string>;
+  requestId?: string;
 }
 
 export const prepareRequest = async (
@@ -114,6 +115,8 @@ export const prepareRequest = async (
     }
   }
 
+  // Propagate under the generic x-request-id header for downstream services
+  // that do not read the snyk-specific header.
   if (payload.headers['snyk-request-id'] && !payload.headers['x-request-id']) {
     payload.headers['x-request-id'] = payload.headers['snyk-request-id'];
   }
@@ -209,6 +212,7 @@ export const prepareRequest = async (
     url: result.url,
     headers: payload.headers,
     method: payload.method,
+    requestId: payload.requestId,
   };
   if (payload.body) {
     req.body = payload.body;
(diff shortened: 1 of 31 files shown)


clwiseman (approved):
Nice work 🥇

PR #1166, opened 2026-06-05
fix(deps): drop uuid@8.3.2 dependency in favor of node:crypto

pavel-snyk:
- [x] Ready for review

#### What does this PR do?
This PR drops usage of uuid@8.3.2 package in favor of builtin `node:crypto`, because we use only `v4`, `validate`, and `NIL`.

The constant name is normalized to use in the same way across the project: `uuidv4`.

`uuid@8.3.2` is flagged by GHSA-w5hq-g745-h8pq (moderate). We don't hit the vulnerable path (v3/v5/v6 + buf), but it's noise in SCA scans. Node >=24 covers `randomUUID` natively, so the package is unnecessary.


Todai88 (commented):
Do we have tests around this to ensure there's no unintentional regression?

Todai88 (approved):

PR #1169, opened 2026-06-08
feat: add typed broker error codes for telemetry

clwiseman:
- [ ] Ready for review

#### What does this PR do?

Add a shared error code catalog so broker failures return consistent code/errorType values and synthesized HTTP statuses only when no downstream response exists. Wire filter blocks, downstream failures, and server logging through the catalog, and stop attempting to update the status code on mid-stream errors after headers are already streamed (it won).

### DIFF
--- a/lib/broker-workload/clientLocalRequests.ts
+++ b/lib/broker-workload/clientLocalRequests.ts
@@ -10,6 +10,10 @@ import {
   WorkloadType,
 } from '../hybrid-sdk/workloadFactory';
 import { ExtendedLogContext } from '../hybrid-sdk/common/types/log';
+import {
+  BROKER_ERROR_CODES,
+  statusForErrorCode,
+} from '../hybrid-sdk/common/types/errorCodes';
 import { incrementHttpRequestsTotal } from '../hybrid-sdk/common/utils/metrics';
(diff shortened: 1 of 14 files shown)


clwiseman:
> [**Runtime Crash Risk**](https://github.com/snyk/broker/pull/1169/changes#diff-ee4966280c6e81bf36b3ef7a063d131f9ac70ac9ddac8d32bb8c04a8e18826deR23-R29) 🟡 [minor]
> The `statusForErrorCode` function throws a hard Error if a `BROKER_ERROR_CODES` value is not present in the `SYNTHESIZED_STATUS` map. Because `SYNTHESIZED_STATUS` is typed as `Partial<Record<BrokerErrorCode, number>>`, the TypeScript compiler will not catch missing mappings at build time. If a developer adds a new error code to the enum but forgets to update this map, the broker will crash with an uncaught exception when that error is encountered. Consider making the map exhaustive (removing `Partial`) or providing a safe default status (e.g., 500) to avoid runtime failures.

This functionality will be extended in the next PR to extend this error catalog to websocket error codes, so I think this an acceptable risk for the current state.

shuaibmunshi (approved):

PR #1171, opened 2026-06-10
fix: make oauth client resilient to transient auth failures

clwiseman:
- [ ] Ready for review

#### What does this PR do?

A single 4xx during auth renewal previously killed the broker client, taking down a healthy client that would have recovered next cycle. This trades aggressive fail-fast for self-healing, cutting avoidable restarts while still surfacing persistent auth problems.

Auth failures now get several chances to refresh a fresh token before giving up; each failure drops the cached token so the next attempt truly re-authenticates, and only sustained failure restarts the process for a clean supervisor relaunch. 401s seen while posting data or reconnecting also force a refresh instead of reusing a rejected token.

### DIFF
--- a/lib/hybrid-sdk/client/metrics/client.ts
+++ b/lib/hybrid-sdk/client/metrics/client.ts
@@ -27,7 +27,11 @@ export interface Client {
    * Best-effort — call forceFlush() if in an async context.
    */
   recordProcessExit(
-    reason: 'reconnect_exhaustion' | 'auth_4xx' | 'uncaught_exception',
+    reason:
+      | 'reconnect_exhaustion'
+      | 'auth_4xx'
+      | 'oauth_token_unavailable'
+      | 'uncaught_exception',
   ): void;
 
   /** Increment broker.client.auth_renewal_failure.total for non-2xx auth renewal responses. */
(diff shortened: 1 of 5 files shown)


clwiseman:
> [**Interface Mismatch**](https://github.com/snyk/broker/pull/1171/changes#diff-4cb22260bb76ed32ac22c76b73cecd57917dafdac59f31f7564c3487e43d1877R29-R35) 🟠 [major]
> Changing the signature of `recordProcessExit` in the `Client` interface to require a `reason` argument breaks the `NoopClient` implementation. The cross-file reference for `lib/hybrid-sdk/client/metrics/noopClient.ts` shows that `NoopClient` still implements `recordProcessExit` with zero parameters. This will result in a TypeScript compilation error.

Tested - this is  incorrect.

> [**Duplicate Error Listeners**](https://github.com/snyk/broker/pull/1171/changes#diff-5f45be3f7a1a2ce506d53ca700d15910ee0ded32832e7a511680b953fc6459b6R363-R366) 🟡 [minor]
> The PR adds a new 'error' listener to the websocket (line 365) to call `invalidateToken()`. However, the existing code (line 362) already attaches the `errorHandler` function to the 'error' event. Depending on the EventEmitter implementation (Primus), attaching multiple listeners is generally safe, but ensure `errorHandler` doesn't have side effects that conflict with an unexpected token invalidation.

This is incorrect. `invalidateToken()` is not a listener it simply nullifies the cached token.

saumilmac (approved):

PR #1179, opened 2026-06-26
feat: client resilience - connection-scoped teardown & response-data fail-fast

skyth3r:
## Summary

Hardens the broker **client** so transient failures don't take down a whole universal-broker pod or hang PR/MR-check requests forever.

This PR introduces 3 changes that are feature flagged (default-on) for now.

## Changes

**1. Connection-scoped auth-renewal teardown** (`BROKER_RENEWAL_CONNECTION_SCOPED_TEARDOWN_ENABLED`)
Auth-renewal exhaustion used to `process.exit(1)`, killing every connection on the pod. Now universal mode tears down only the failing connection's socket pair (clearing both sockets' renewal timers); classic mode still exits and is not affected by this change. 

For metrics the `client-shutdown`  metric is only on the classic path now; a new `connection_teardown` metric is emitted otherwise.

**2. Fail fast on failed response-data uploads** (`BROKER_RESPONSE_DATA_FAILFAST_ENABLED`)
A failed/stalled upload (e.g. edge `503`) used to hang the request until a far-off platform timeout. Now the server arms an active per-stream deadline (`RESPONSE_DATA_TIMEOUT_MS`, default 30m) and the client emits a WebSocket `abort` on failure, so the request fails fast with a `5xx`.

**3. Drop a no-op response-error retry in `makeStreamingRequestToDownstream`**
`resolve(response)` already settled the Promise, so the old `response.on('error')` retry was a no-op that spawned a dangling Promise (unhandled rejection). Mid-stream errors are already handled by `forwardRequest`'s own handler.

## Testing
Unit tests per path + a functional test proving the request fails fast (`5xx` in <5s, not hung). `build`, `lint`, and the full unit suite pass as well.


skyth3r:
### Addressing comments from review bot

**Partial Teardown Logic (major) - addressed in `d1d3d80`.**

The `connectionTornDown = true` decision is correct here (we only reach this branch after `MAX_CONSECUTIVE_AUTH_FAILURES` consecutive renewal failures, so re-arming renewal would just loop-and-fail). 

The `idx === -1` branch now closes the current socket directly (clears its
renewal timer, then `end()`/`destroy()`) instead of only logging, so every
teardown path ends with an actual clean close. Added a unit test covering the
not-in-registry case.

**Immediate Socket Destruction (minor) - leaving as-is**

This `end()` → `destroy()` sequence is pre-existing behaviour: it already lived in `shutDownConnectionPair`, and this PR's refactor only extracted it into the `clearAndStop` helper without changing the sequence.

**Possible Crash (major)**

`shutDownConnectionPair` now guards both indexes: it returns early if the primary index points at nothing, and only tears down the sibling when `secondTunnelIndex !== -1`. 

So a half-removed pair (or a `-1` index from a caller) results in a safe no-op instead of a `TypeError` that would crash the client. Unit tests to check for the missing-sibling and out-of-range-index cases have been added.

lib/hybrid-sdk/client/socket.ts
```diff
@@ -273,18 +277,59 @@ export const createWebSocket = (
         consecutiveAuthFailures++;
         metricsClient.recordAuthRenewalFailure(statusCode);
 
+        const teardownEnabled =
+          String(
+            clientOpts.config.brokerRenewalConnectionScopedTeardownEnabled,
+          ) !== 'false';
```
mikedevelops, line 283: Why are we coercing this to string to do a comparison? Does the way we load the config make this necessary?

lib/hybrid-sdk/client/socket.ts
```diff
@@ -273,18 +277,59 @@ export const createWebSocket = (
...
+
         if (consecutiveAuthFailures >= MAX_CONSECUTIVE_AUTH_FAILURES) {
+          if (clientOpts.config.universalBrokerEnabled && teardownEnabled) {
+            // Universal mode: drop ONLY this connection's pair; keep the process (and
+            // every other connection) alive.
```
mikedevelops, line 288: What is the fate of the connections that have been dropped in this scenario? Previously we would stop the process so all connections would be re-established on startup (presumably). In this new world could we end up in a scenario where a broker client appears to be healthy but only _some_ of the configured connections are established to the server?

PR #1181, opened 2026-07-09
feat: connection-scoped auth-renewal teardown with self-healing

skyth3r:
## Summary

Previously, when a broker **client** exhausted auth-renewal retries (3 consecutive failures), it called `process.exit(1)` — killing the whole pod and, in universal mode, **every** connection on it, not just the failing one.

This PR makes universal mode tear down **only the failing connection's socket pair** and keep the process (and all other connections) alive, then **self-heal** that connection. If self-healing can't recover it, the connection becomes **visible as unhealthy** rather than silently absent. Classic mode is unchanged — it still exits so its supervisor restarts it.

All behavior is feature-flagged (default-on).

## Background — the gap this closes

Scoped teardown on its own isn't enough: a torn-down connection was never re-established in Universal GA steady state (polling is disabled; the backup watcher only fires when *zero* connections are open), and `/healthcheck` still returned `200` because the dead pair was spliced out of the registry. So a client could look healthy while silently missing a configured connection. This PR closes that gap.

## Changes

**1. Connection-scoped teardown** (`BROKER_RENEWAL_CONNECTION_SCOPED_TEARDOWN_ENABLED`)
Universal mode clears both sockets' renewal timers and tears down just the failing pair. The `client-shutdown` metric is now classic-only; universal emits a new `connection_teardown` metric.

**2. Self-healing re-establishment** (`BROKER_RENEWAL_REESTABLISHMENT_ENABLED`)
After teardown, the connection is re-established on a **capped exponential backoff** (`base·2^attempt`, base 60s, default **3** attempts) that delegates to the existing `syncClientConfig` reconciler — no duplicated connection-creation logic. Because teardown means auth is failing, we never re-create immediately. If a re-created connection renews successfully, its backoff budget resets. Once the cap is exhausted, the connection is marked *gave-up*.

**3. Health visibility**
`/healthcheck` now diffs configured vs. established connections and reports any missing one in the body (`degraded: true`, `reestablishment: 'reestablishing' | 'gave_up'`):
- **still retrying** → HTTP **200** (degraded body) — don't restart the pod mid-recovery; other connections stay up.
- **gave up** → HTTP **500** — the probe restarts the pod so all connections re-establish fresh.

New `connection_reestablishment` metric (`attempt` / `success` / `exhausted`) and a `friendly_name` label on the teardown metric localize failures per connection.

## Escalation model

Graceful-first, restart-as-last-resort: scoped teardown → backed-off self-heal → if that's exhausted, surface `500` so the platform restarts the pod. This keeps the "don't nuke the pod on a transient blip" goal while guaranteeing a permanently-dead connection can't hide behind a green healthcheck.

## Testing

- **Unit**: new `reestablishment` suite (backoff, cap→gave-up, shutdown short-circuit, budget reset, reschedule-on-error); extended healthcheck (degraded/200 + gave-up/500 + skips disabled/identifier-less), socket wiring (schedule on teardown, clear on renewal), and metrics (`friendly_name` + reestablishment counter).
- **Functional**: universal `/healthcheck` asserts a torn-down connection is surfaced as `degraded`/`reestablishing` while the endpoint stays `200`.
- `build`, `lint`, full unit suite, and the functional suite all pass.

## Feature flags (all default-on)

| Flag | Default | Purpose |
|---|---|---|
| `BROKER_RENEWAL_CONNECTION_SCOPED_TEARDOWN_ENABLED` | `true` | Universal: teardown instead of process exit |
| `BROKER_RENEWAL_REESTABLISHMENT_ENABLED` | `true` | Self-heal torn-down connections |
| `BROKER_RENEWAL_MAX_REESTABLISHMENT_ATTEMPTS` | `3` | Backoff attempt cap before giving up |


skyth3r:
> [**Object methods on Map**](https://github.com/snyk/broker/pull/1181/changes#diff-7947396a726b90ea24a12df6f806e089521e99dc9855acad55e39b9708fc2e26R61-R61) 🟠 [major]
> The `healthCheckHandler` attempts to iterate over configured connections using `Object.keys(configuredConnections)`. However, in universal mode, the connection configuration is stored in a `Map` (as seen in `systemCheckHandler.ts` and `synchronizer.ts`). Calling `Object.keys` on a `Map` will return an empty array, causing the healthcheck to fail to report any missing or unhealthy connections in universal mode.
> 
> ```ts
> for (const friendlyName of Object.keys(configuredConnections)) {
> ```

This is a false positive - config.connections is a plain object (Record<string, any>), not a Map

This is likely a confusion that stems from the derived Map that getConfigForConnections() builds from the plain object (universal.ts:33-37) and which systemCheckHandler.ts:23 iterates. They are two different values. 

The healthcheck reads config.connections (the object).

> [**Logic Error**](https://github.com/snyk/broker/pull/1181/changes#diff-93a526526c8adb655c89b48ecaade479838ca750cedeca900e39f40cb0eff5e7R60-R64) 🟡 [minor]
> In `clearReestablishmentState`, the function deletes the timer from `timersByConnection` after calling `clearTimeout`, but then checks `wasRecovering` (calculated using `getReestablishmentState`) to determine whether to log info and record metrics. Because `getReestablishmentState` checks the maps that were just mutated, it might return inconsistent results if the state was solely defined by a pending timer.
> 
> ```ts
>   clearTimeout(timer);
>   timersByConnection.delete(friendlyName);
> }
> attemptsByConnection.delete(friendlyName);
> gaveUpConnections.delete(friendlyName);
> ```

This is also incorrect. 

wasRecovering is captured **before** any map is mutated.
getReestablishmentState already checks timersByConnection.has(...) (lines 37-39), so the pending-timer-only state is covered

PR #1209, opened 2026-08-17
fix: cache compiled regexes in filter and log-redaction hot paths

clwiseman:
- [ ] Ready for review
- [ ] Follows [CONTRIBUTING](https://github.com/snyk/broker/blob/main/.github/CONTRIBUTING.md) rules

#### What does this PR do?

Compile each RegExp object once at startup rather than on every request/log call. Adds test coverage for the changed body-regex path and for previously-untested universal-broker log redaction.

### DIFF
--- a/lib/hybrid-sdk/common/filter/filtersAsync.ts
+++ b/lib/hybrid-sdk/common/filter/filtersAsync.ts
@@ -95,9 +95,20 @@ export const loadFilters: LOADEDFILTER = (
     method = (method || 'get').toLowerCase();
 
     const bodyFilters = valid ? valid.filter((v) => !!v.path && !v.regex) : [];
-    const bodyRegexFilters = valid
-      ? valid.filter((v) => !!v.path && !!v.regex)
-      : [];
+    const bodyRegexFilters = (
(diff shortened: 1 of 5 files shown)


clwiseman:
> [**Memory Leak**](https://github.com/snyk/broker/pull/1209/changes#diff-312bf437521d1249c771663d9d603191e2bf7648e236dcdeb657bc365affadb0R24-R33) 🟠 [major]
> The `compiledRedactionRegexes` Map is an unbounded global cache. In environments where credentials (like `GITHUB_TOKEN` or connection identifiers) rotate frequently or where a large number of unique identifiers are logged, this Map will grow indefinitely. Since entries are never evicted, this will lead to a slow memory leak and eventual OOM in long-running processes. A size-limited cache (LRU) or clearing the cache on config reload should be used instead.

This is a theoretical issue but not likely at all to occur in reality. The client would have to have thousands of unique credentials requiring redactions (or be constantly live-rotating tokens) for this to become an issue, which just wouldn't happen in practice. Especially since the cache is cleared if the container restarts.

shuaibmunshi (approved):

PR #1211, opened 2026-08-20
fix: broker server stale connections [ACC-3561]

shuaibmunshi:
- [x] Ready for review
- [x] Follows [CONTRIBUTING](https://github.com/snyk/broker/blob/main/.github/CONTRIBUTING.md) rules

#### What does this PR do?
Adds a 30 second prune timer for connections that initiate but never finish the handshake. Previously these incomplete connections would remain in memory indefinitely and return 503 errors unnecessarily. 

Also fixes a type error if a connection in the auth handler does not have metadata.

### DIFF
--- a/lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
+++ b/lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
@@ -10,6 +10,74 @@ import { makeStreamingRequestToDownstream } from '../../http/request';
 import { retry } from '../../http/exponential-backoff';
 
 const CONNECTION_READY_MAX_RETRIES = 4;
+const CONNECTION_HANDSHAKE_TTL_MS = Number(
+  process.env.BROKER_CONNECTION_HANDSHAKE_TTL_MS ?? 30_000,
+);
+
+const isReady = (client: ClientSocket): boolean =>
+  Boolean(
+    client.socket && client.metadata?.version && client.metadata?.capabilities,
+  );
+
+/**
+ * Drops authorization placeholders that never completed the identify handshake.
+ * Ready connections and legacy entries without a handshake timestamp are kept.
+ */
+const removeExpiredPendingHandshakes = (
+  connections: Map<string, ClientSocket[]>,
+  token: string,
+  now = Date.now(),
+): void => {
+  const activePool = (connections.get(token) ?? []).filter((client) => {
+    if (isReady(client)) {
+      return true;
+    }
+    // Only authorize-only entries are pending. An identified socket with
+    // incomplete metadata must remain tracked and follow the bad-metadata path.
+    if (
+      client.socket ||
+      client.metadata ||
+      client.handshakeStartTime === undefined
+    ) {
+      return true;
+    }
+    return now - client.handshakeStartTime < CONNECTION_HANDSHAKE_TTL_MS;
+  });
+
+  if (activePool.length > 0) {
+    connections.set(token, activePool);
+  } else {
+    connections.delete(token);
+  }
+};
+
+// The pool is newest-first, so find() returns the newest fully-ready connection.
+// This keeps HA clients immune to reconnect churn, which briefly creates a
+// not-yet-ready connection.
+type ConnectionLookup =
+  | { status: 'ready'; client: ClientSocket }
+  | { status: 'no-connection' }
+  | { status: 'bad-metadata' };
+
+const findReadyConnection = (
+  connections: Map<string, ClientSocket[]>,
+  token: string,
+): ConnectionLookup => {
+  removeExpiredPendingHandshakes(connections, token);
+  const pool = connections.get(token) ?? [];
+  if (pool.length === 0) {
+    return { status: 'no-connection' };
+  }
+  const ready = pool.find(isReady);
+  if (ready) {
+    return { status: 'ready', client: ready };
+  }
+  if (pool.some((client) => !client.socket || !client.metadata)) {
+    // Still completing the handshake. Throwing asks retry() for another attempt.
+    throw new Error('client handshake in progress');
+  }
+  return { status: 'bad-metadata' };
+};
 
 export const overloadHttpRequestWithConnectionDetailsMiddleware = async (
   req: Request,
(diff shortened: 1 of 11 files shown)


lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
```diff
@@ -77,9 +108,10 @@ export const overloadHttpRequestWithConnectionDetailsMiddleware = async (
       return res.status(404).json({ ok: false });
     }
   }
-  // Grab the pool of connections for this token
-  const connection = connections.get(token)!;
-  // Make sure a connection actually exists before proceeding
+  // Remove authorization placeholders that never completed the identify
+  // handshake. Ready connections and legacy entries without a handshake
+  // timestamp are preserved.
+  const connection = pruneExpiredHandshakes(connections, token);
```
pavel-snyk: @shuaibmunshi, is my understanding correct here?
1. Cleanup only runs when broker/service traffic arrives for this token. Without a subsequent request, the stale entry remains in the global map and status endpoints can continue reporting it. Can expiration be owned by the connection lifecycle instead of this request middleware?
2. `pruneExpiredHandshakes` helper both mutates the global map and returns the replacement pool, which makes the ownership and freshness of the returned value unclear. The call site also names the array connection. How about making this a void command:

```
removeExpiredPendingHandshakes(connections, token);
const pool = connections.get(token) ?? [];
```

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
```diff
@@ -100,7 +130,7 @@ export const overloadHttpRequestWithConnectionDetailsMiddleware = async (
   try {
     client = await retry<ClientSocket | undefined>(
       () => {
-        const pool = connections.get(token) ?? [];
+        const pool = pruneExpiredHandshakes(connections, token);
```
pavel-snyk: If the final handshake expires during retry backoff, this returns an empty pool. The callback then returns undefined, and the handler reports 400 bad-request even though there is no connection. Can we handle the empty-pool state explicitly and add a TTL-crossing test please?

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
shuaibmunshi: Yeah good point, I made it run on socket close and left the TTL as a backup. The helper is now a void.

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
shuaibmunshi: Yeah thanks I made it a 404 like the others.

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
shuaibmunshi: Yeah good point, its not on socket close.

The helper is now void.

lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
```diff
@@ -38,13 +39,19 @@ export const handleConnectionCloseOnSocket = (
...
       decrementSocketConnectionGauge();
+      rmClientIdFromTerminationMap(token, clientId);
+      setImmediate(async () => await clientDisconnected(token, clientId));
     } else {
+      // No pool entry references this socket, because the socket is only
+      // attached by identify. The authorize hook still seated an entry for this
+      // handshake, and nothing else removes it. Without this the entry remains
+      // until an inbound request prunes it by handshake TTL.
+      const identity = getHandshakeIdentityFromHeaders(socket.request.headers);
+      const removed = removePendingHandshake(token, identity);
```
pavel-snyk, line 50: @shuaibmunshi, handshakeId does not fully prevent an older connection from removing a newer reconnect.

e.g. if A is identified, authorize(B) merges B’s handshake/auth fields into A’s entry while retaining socket A. If A closes before B identifies, the identified cleanup removes the entry. B then identifies into a fresh entry without `brokerAppClientId`, `role`, or `credsValidationTime`, so auth refresh cannot match it and the stale-creds sweep will eventually close it.

Could we preserve B’s pending state when the closing socket’s handshakeId differs from the entry’s, or model each handshake separately, and add this ordering as a regression test?

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
```diff
@@ -10,6 +10,68 @@ import { makeStreamingRequestToDownstream } from '../../http/request';
...
+  now = Date.now(),
+): void => {
+  const activePool = (connections.get(token) ?? []).filter((client) => {
+    if (isReady(client)) {
+      return true;
+    }
+    if (client.handshakeStartTime === undefined) {
+      return true;
+    }
+    return now - client.handshakeStartTime < CONNECTION_HANDSHAKE_TTL_MS;
```
pavel-snyk, line 44: @shuaibmunshi, this filters every non-ready timestamped entry, not only authorize-only placeholders.

An identified socket with metadata but missing capabilities retains `handshakeStartTime`; after the TTL it is removed instead of continuing through the intended bad-metadata/400 path, and the live socket becomes untracked.

Could expiration use an explicit pending state - or at least require the socket/metadata to be absent - and add an expired identified-socket test?

pavel-snyk (changes requested):
@shuaibmunshi, thanks for working through the feedback. I like the direction and the updates:

  - `metadata.clientId` -> `brokerClientId`
  - generic 500 response with underlying error logged
  - void pruning helper
  - ttl expiration during retry now returns 404, with a regression test

There are just two remaining connection-lifecycle edge cases in the inline comments. Once those are addressed, we should be good to merge.

Also, please squash the commits before merging-it will keep the release notes much cleaner.

lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
shuaibmunshi: Thanks, fixed

lib/hybrid-sdk/server/routesHandlers/httpRequestHandler.ts
shuaibmunshi: thanks fixed

pavel-snyk (approved):
LGTM, some non-blocking findings can be addressed with the next PR:
  - validate BROKER_CONNECTION_HANDSHAKE_TTL_MS eventually.
  - fix/remove the misleading clientDisconnected assertion in `test/unit/lib/hybrid-sdk/server/socketHandlerscloseHandler.test.ts`, since it asserts before the scheduled setImmediate runs

PR #1212, opened 2026-08-25
fix: return retryable response during broker reconnect

skyth3r:
## Summary

When a Broker client's last live websocket connection to a Broker server instance closes, routing can briefly continue sending requests to that instance while the client reconnects. Requests arriving during this window find no usable local connection and receive a `404`, incorrectly representing a temporary connection gap as a permanent failure.

This PR fixes this by returning a retryable `503` during this window, signalling that the failure is temporary. It also handles reconnects that are authorized but close before identifying by distinguishing pending handshakes from live websocket connections.

The overall flow is now this:

```
 Last live socket closes
     ↓
 Active socket is removed
     ↓
 Token is marked as recently disconnected for 60 seconds,
 even if a pending handshake remains
     ↓
 Request arrives without a usable connection
     ├─ Token was recently disconnected or is still reconnecting → Return 503 (safe to retry)
     └─ Token is unknown or the window expired → Return the existing 404
```

This distinction prevents a temporary reconnect window, including a reconnect that fails before identifying, from looking like a permanently unknown Broker token.

## Changes made

- Track recently disconnected Broker tokens for 60 seconds.
- Distinguish live websocket connections from authorized but unidentified pending handshakes.
- Retain the disconnect marker when the last live socket closes, even if a pending handshake remains.
- Return `503` with `Retry-After: 1` and `x-broker-failure: connection-not-ready` during the reconnect window.
- Preserve the existing `404 no-connection` response for unknown or long-disconnected tokens.
- Preserve secondary-to-primary request forwarding.
- Add unit and functional coverage for disconnect expiry, repeated disconnects, failed pending reconnects, surviving live sockets, forwarding, and the retryable response.

### DIFF
--- a/lib/hybrid-sdk/server/socket.ts
+++ b/lib/hybrid-sdk/server/socket.ts
@@ -33,9 +33,11 @@ const socketConnections = new Map<string, ClientSocket[]>();
 const RECONNECT_GRACE_MS = 60_000;
 const recentlyDisconnectedExpiryTimers = new Map<string, NodeJS.Timeout>();
 
-export const removeSocketConnection = (token: string) => {
-  socketConnections.delete(token);
-
+/**
+ * Starts or refreshes the reconnect grace period without changing the pool.
+ * A token can still have authorized handshakes even when it has no live socket.
+ */
+export const markRecentlyDisconnected = (token: string) => {
   const previousExpiryTimer = recentlyDisconnectedExpiryTimers.get(token);
   if (previousExpiryTimer) {
     clearTimeout(previousExpiryTimer);
@@ -48,6 +50,11 @@ export const removeSocketConnection = (token: string) => {
   }, RECONNECT_GRACE_MS);
   expiryTimer.unref();
   recentlyDisconnectedExpiryTimers.set(token, expiryTimer);
+};
+
+export const removeSocketConnection = (token: string) => {
+  socketConnections.delete(token);
+  markRecentlyDisconnected(token);
 };
 
 export const wasRecentlyDisconnected = (token: string): boolean =>
--- a/lib/hybrid-sdk/server/socket.ts
+++ b/lib/hybrid-sdk/server/socket.ts
@@ -30,6 +30,28 @@ export interface ClientSocket {
   handshakeId?: string;
 }
 const socketConnections = new Map<string, ClientSocket[]>();
+const RECONNECT_GRACE_MS = 60_000;
+const recentlyDisconnectedExpiryTimers = new Map<string, NodeJS.Timeout>();
+
+export const removeSocketConnection = (token: string) => {
+  socketConnections.delete(token);
+
+  const previousExpiryTimer = recentlyDisconnectedExpiryTimers.get(token);
+  if (previousExpiryTimer) {
+    clearTimeout(previousExpiryTimer);
+  }
+
+  const expiryTimer = setTimeout(() => {
+    if (recentlyDisconnectedExpiryTimers.get(token) === expiryTimer) {
+      recentlyDisconnectedExpiryTimers.delete(token);
+    }
+  }, RECONNECT_GRACE_MS);
+  expiryTimer.unref();
+  recentlyDisconnectedExpiryTimers.set(token, expiryTimer);
+};
+
+export const wasRecentlyDisconnected = (token: string): boolean =>
+  recentlyDisconnectedExpiryTimers.has(token);
 
 export const getSocketConnections = () => {
   return socketConnections;
--- a/lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
+++ b/lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
@@ -1,7 +1,11 @@
 import { decrementSocketConnectionGauge } from '../../common/utils/metrics';
 import { log as logger } from '../../../logs/logger';
 import { clientDisconnected } from '../infra/dispatcher';
-import { getSocketConnections, removePendingHandshake } from '../socket';
+import {
+  getSocketConnections,
+  removePendingHandshake,
+  removeSocketConnection,
+} from '../socket';
 import { getHandshakeIdentityFromHeaders } from '../auth/authHelpers';
 import { getDesensitizedToken } from '../utils/token';
 import { rmClientIdFromTerminationMap } from './identifyHandler';
@@ -36,7 +40,7 @@ export const handleConnectionCloseOnSocket = (
         connections.set(token, filteredClientPool);
       } else {
         logger.info({ maskedToken, hashedToken }, 'Removing client.');
-        connections.delete(token);
+        removeSocketConnection(token);
       }
       decrementSocketConnectionGauge();
       rmClientIdFromTerminationMap(token, clientId);
--- a/lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
+++ b/lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
@@ -3,6 +3,7 @@ import { log as logger } from '../../../logs/logger';
 import { clientDisconnected } from '../infra/dispatcher';
 import {
   getSocketConnections,
+  markRecentlyDisconnected,
   removePendingHandshake,
   removeSocketConnection,
 } from '../socket';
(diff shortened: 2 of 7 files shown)


lib/hybrid-sdk/server/socket.ts
```diff
@@ -26,6 +26,28 @@ export interface ClientSocket {
   credsValidationTime?: string;
 }
 const socketConnections = new Map<string, ClientSocket[]>();
+const RECONNECT_GRACE_MS = 60_000;
+const recentlyDisconnectedExpiryTimers = new Map<string, NodeJS.Timeout>();
+
+export const removeSocketConnection = (token: string) => {
```
skyth3r: New socket function - This does two things:
1. Removes a token from the active socketConnections map
2. Starts a 60-second grace period (set by `RECONNECT_GRACE_MS`) by storing an expiry timer for that token

lib/hybrid-sdk/server/socket.ts
```diff
@@ -26,6 +26,28 @@ export interface ClientSocket {
...
+  if (previousExpiryTimer) {
+    clearTimeout(previousExpiryTimer);
+  }
+
+  const expiryTimer = setTimeout(() => {
+    if (recentlyDisconnectedExpiryTimers.get(token) === expiryTimer) {
+      recentlyDisconnectedExpiryTimers.delete(token);
+    }
+  }, RECONNECT_GRACE_MS);
+  expiryTimer.unref();
```
skyth3r, line 51: Unreferenced with .`unref()`, to prevent disruptions during shutdown

lib/hybrid-sdk/server/socket.ts
```diff
@@ -26,6 +26,28 @@ export interface ClientSocket {
...
+  recentlyDisconnectedExpiryTimers.set(token, expiryTimer);
+};
+
+export const wasRecentlyDisconnected = (token: string): boolean =>
```
skyth3r, line 60: New socket function - Checks if token has an active expiry timer

skyth3r:
>[Memory Exhaustion Risk](https://github.com/snyk/broker/pull/1212/changes#diff-987c4218f7dff0905998e97c0111234a84b6916f1a56d5735ea18a1dffaa06b6R30-R30) 🟠 [major]
The recentlyDisconnectedExpiryTimers Map lacks a maximum size limit. In a scenario with a high volume of unique tokens connecting and disconnecting (e.g., a churn event or a DoS attempt), this map could grow significantly, consuming memory for the duration of the 60-second RECONNECT_GRACE_MS window. Consider implementing a maximum size or an LRU policy for this map.

Non-issue. This map is time-bound and cannot be populated with arbitrary HTTP requests and entries are removed after 60 seconds.

The existing `socketConnections` map also has no maximum-size limit and holds more state for every active client. 
The new map briefly retains only the token and expiry timer after that larger connection state has been removed. Its size is bound by the number of unique final client disconnects on this process within a 60-second window.

skyth3r:
>[Stale Disconnect Status](https://github.com/snyk/broker/pull/1212/changes#diff-987c4218f7dff0905998e97c0111234a84b6916f1a56d5735ea18a1dffaa06b6R49-R50) 🟡 [minor]
There is no logic to clear a token from recentlyDisconnectedExpiryTimers when a client successfully reconnects. If a client reconnects 5 seconds into the 60-second grace period, wasRecentlyDisconnected(token) will continue to return true for the remaining 55 seconds. While httpRequestHandler.ts currently mitigates this by checking connections.has(token) first, this creates a state inconsistency that could affect future contributors or other modules relying on the 'recently disconnected' flag.

While true, the marker intentially represents recent disconnect history rather than the client's current connection status.

If the token reconnects during the grace period, it is inserted back into `socketConnections`. The request handler checks that map first and so `wasRecentlyDisconnected` is *not* checked when there is an active socket connection. Thus a retained marker in this map cannot produce a 503 while a client is connected, and is removed automatically when the 60-second window expires. 

The function is called `wasRecentlyDisconnected` and whether the token reconnects or not, it still signifies that a token was historically disconnected. I don't this this adds much at this moment in time, but neither does the removal of a reconnected token from the map. Currently there is currently no other logic that relies on this 'recently disconnected' logic.In the future we can always come back to this and improve it if needed.

saumilmac (dismissed):

pavel-snyk (changes requested):
@skyth3r, solid, the distinction between an unknown token (404) and a temporarily disconnected broker (503) makes sense.

There is one reconnect sequence that I think we should handle before merging:

1. identified broker connection exists.
2. reconnect is authorized, creating a pending pool entry without a socket.
3. the old identified socket closes. `closeHandler.ts` correctly preserves the pending entry.
4. the pending reconnect also closes before identifying.
5. `removePendingHandshake()` deletes the now-empty pool directly.
6. because the recently-disconnected marker was never set, the next request returns 404 instead of the intended retryable 503.

The key distinction is that "the pool contains an entry" does not necessarily mean "the token has a live Broker connection." Pending handshakes are stored without a socket.

lib/hybrid-sdk/server/socket.ts
```diff
@@ -30,6 +30,28 @@ export interface ClientSocket {
   handshakeId?: string;
 }
 const socketConnections = new Map<string, ClientSocket[]>();
+const RECONNECT_GRACE_MS = 60_000;
+const recentlyDisconnectedExpiryTimers = new Map<string, NodeJS.Timeout>();
+
+export const removeSocketConnection = (token: string) => {
```
pavel-snyk: extracting marker logic in pseudocode (something like that)

```
function markRecentlyDisconnected(token):
      cancel any existing expiry timer for token

      timer = start 60-second timer:
          remove recently-disconnected marker for token

      timer.unref()
      store timer for token

function removeSocketConnection(token):
      delete token from socketConnections
      markRecentlyDisconnected(token)
```

lib/hybrid-sdk/server/socketHandlers/closeHandler.ts
```diff
@@ -36,7 +40,7 @@ export const handleConnectionCloseOnSocket = (
         connections.set(token, filteredClientPool);
       } else {
         logger.info({ maskedToken, hashedToken }, 'Removing client.');
-        connections.delete(token);
+        removeSocketConnection(token);
```
pavel-snyk, line 44: applying the live-socket check (pseudo-code):

```
remainingConnections = pool excluding closing socket

if remainingConnections is empty:
    removeSocketConnection(token)
else:
    connections.set(token, remainingConnections)

    hasLiveSocket = remainingConnections.some(
        connection => connection.socket exists
    )

    if not hasLiveSocket:
        // Preserve the pending handshake, but record that there is
        // currently no usable Broker connection.
        markRecentlyDisconnected(token)
```

skyth3r:
>[Orphaned Timers](https://github.com/snyk/broker/pull/1212/changes#diff-987c4218f7dff0905998e97c0111234a84b6916f1a56d5735ea18a1dffaa06b6R36-R51) 🟡 [minor]
The recentlyDisconnectedExpiryTimers for a token is not cleared when a broker client successfully reconnects and identifies. While overloadHttpRequestWithConnectionDetailsMiddleware guards against this by checking connections.has(token) first, the timer object remains in the Node.js event loop for the full 60 seconds even after a healthy connection is restored. In environments with high connection churn, this could lead to a large number of unnecessary pending timeouts.

Incorrect. The timeout is `unref()`'d in `lib/hybrid-sdk/server/socket.ts:L49`, so it does not keep the Node.js event loop or process alive. Keeping the timer object for the full 60 seconds even after a healthy connection is restored is expected.

skyth3r:
>[Improper Retry Signaling](https://github.com/snyk/broker/pull/1212/changes#diff-5a6ab885e50b100dd4ba14208b0c3ab696151f1ecfca5ac23fb705fa0522bd0cR40-R40) 🟠 [major]
Graceful shutdowns (e.g. via SIGTERM) trigger markRecentlyDisconnected(token) regardless of intent. The system already tracks intentional shutdowns in terminatingClientIdsPerToken (see cross_file_context: identifyHandler.ts). By returning a 503 'Safe to retry' during a graceful shutdown, the Snyk Platform will unnecessarily retry for 60 seconds against a client that is permanently gone, delaying failover or clean failure reporting.

Accepted for now - Can be addressed in a follow up PR if needed

pavel-snyk (dismissed):
Thanks for the quick iteration. The updated implementation addresses the reconnect race we discussed: it preserves pending handshakes while correctly marking the token when no live socket remains. The new regression tests cover both the failed reconnect and the multiple-live-socket case well.

We noticed one related TTL cleanup path where an expired pending handshake can still result in a 404 while the 60-second reconnect marker is active. We can address that separately in the next PR, it is not blocking this one.

Nice work. Approved from my side 🚀

pavel-snyk (approved):

PR #1220, opened 2026-09-06
refactor: modernize pending POST stream response delivery

pavel-snyk:
- [ ] Ready for review
- [ ] Follows [CONTRIBUTING](https://github.com/snyk/broker/blob/main/.github/CONTRIBUTING.md) rules

### What does this PR do?
This spike modernizes how Broker Server delivers streamed responses posted back by Broker Client.

Previously, pending responses, metadata decoding, stream buffering, ownership checks, cleanup, and metrics were spread across several handlers. The POST path also manually consumed `data` events and coordinated backpressure around an intermediate `PassThrough`.

This change introduces explicit framing and pending-response abstractions, then uses Node.js pipelines to connect response sources to the original HTTP response.

```
  Original HTTP request
          │
          ▼
  PendingResponseRegistry.register()
          │
          ├── POST delivery
          │     request → ResponseFrameDecoder → original HTTP response
          │
          └── legacy WebSocket delivery
                callback chunks → PassThrough → original HTTP response
```

Only one delivery path can claim a pending response.

#### Response frame decoder

Introduces `ResponseFrameDecoder`, which decodes the existing response format:

> `4-byte little-endian metadata length | JSON metadata | body bytes`

The decoder:

  - handles fragmented length prefixes
  - handles metadata split across chunks and UTF-8 code points
  - applies metadata before forwarding body bytes
  - streams body bytes without waiting for source EOF
  - participates in pipeline backpressure
  - Rejects incomplete or malformed framing
  - validates response status, headers, and errorType
  - applies a defensive 1 MiB metadata limit

The wire format itself is unchanged.

#### Pending-response registry

Replaces the shared stream-store representation with `PendingResponseRegistry`.

The registry:

  - registers the original HTTP response as the destination
  - grants a one-shot claim to either POST or legacy WebSocket delivery
  - verifies Broker or legacy connection ownership before consuming a claim
  - distinguishes missing, already-claimed, and owner-mismatch outcomes
  - rejects duplicate registrations
  - handles requester disconnects and response errors
  - expires unclaimed registrations
  - allows an already-claimed delivery to complete after its registration TTL
  - makes completion and cancellation idempotent
  - records response-size metrics after successful completion

#### POST response delivery

`handlePostResponse()` now uses:

> `POST request → ResponseFrameDecoder → original HTTP response`

It:

  - claims the pending response before consuming the POST body
  - applies decoded status and headers to the original response
  - waits for final destination delivery before acknowledging the POST
  - returns `200 {}` only after successful delivery
  - cancels the claim on source, framing, metadata, or destination failure
  - preserves JWT azp ownership validation

#### Legacy WebSocket compatibility

The legacy callback-based response path now uses the same pending-response registry.

A `PassThrough` adapts WebSocket chunks into a pipeline while preserving:

  - the existing callback protocol
  - response metadata application
  - response-size metrics
  - legacy connection ownership
  - protection against POST and WebSocket delivery claiming the same response

The legacy protocol does not provide a drain acknowledgement, so its existing buffering limitation remains.

#### Request registration

The WebSocket request path now registers the original HTTP response directly.

If the requester has already disconnected, registration returns destination-unavailable and no WebSocket request is sent.

### DIFF
--- a/lib/hybrid-sdk/http/server-post-stream-handler.ts
+++ b/lib/hybrid-sdk/http/server-post-stream-handler.ts
@@ -1,79 +1,289 @@
-import stream from 'stream';
-
+import { Writable } from 'node:stream';
 import { Response } from 'express';
 import NodeCache from 'node-cache';
 import { getConfig } from '../common/config/config';
 import { observeResponseSize } from '../common/utils/metrics';
+import { ResponseMetadata } from './response-frame-decoder';
 
-export const streamsStore = new NodeCache({
-  stdTTL: parseInt(getConfig().cacheExpiry) || 3600, // 1 hour
-  checkperiod: parseInt(getConfig().cacheCheckPeriod) || 60, // 1 min
-  useClones: false,
-});
+export interface PendingResponseRegistration {
+  readonly response: Response;
+  readonly brokerAppClientId: string | null;
+  readonly legacyConnectionIdentifier?: string;
+}
 
-export interface StreamResponse {
-  streamBuffer: stream.PassThrough;
+export interface ClaimOptions {
+  brokerAppClientId?: string;
+  enforceBrokerOwnership?: boolean;
+  legacyConnectionIdentifier?: string;
+  enforceLegacyOwnership?: boolean;
+}
+
+export type RegistrationResult =
+  | { status: 'registered' }
+  | { status: 'destination-unavailable' };
+
+export type ClaimResult =
+  | { status: 'claimed'; pending: PendingResponse }
+  | { status: 'not-found' | 'already-claimed' | 'owner-mismatch' };
+
+type Settlement =
+  | { type: 'complete'; bodyBytes: number }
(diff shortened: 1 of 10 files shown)


lib/hybrid-sdk/http/server-post-stream-handler.ts
```diff
@@ -1,79 +1,289 @@
...
     });
-  };
+    return { status: 'registered' };
+  }
 
-  destroy = (error) => {
-    this.streamResponse.streamBuffer.destroy(error);
-    streamsStore.del(this.streamingID);
-  };
+  claim(streamingID: string, options: ClaimOptions = {}): ClaimResult {
```
shuaibmunshi: Nit: Does it make sense to have a TTL on these claims or to flush them when a websocket is cancelled? I think in the current case a claim for a severed websocket might remain in activeClaims longer than necessary, is that the case?

shuaibmunshi (approved):

